PR #5, opened 2023-05-23
landing was created

Yuliia-Hrinchenko1:
[DEMO LINK](https://Yuliia-Hrinchenko1.github.io/layout_landing-page/)


vlad-demchuk (changes requested):
1. phone link leads to 404 error and mail leads just to another block, however it should be mailto and tel links
image
2. check these checklist points
image

nazarmatsevych (approved):

PR #11, opened 2023-05-25
Landing page

MaxSchmide:
- [DEMO LINK](https://maxschmide.github.io/layout_landing/)


mgrinko (changes requested):
Backgrounds should not be limited on large screens

mgrinko (changes requested):
First screen does not look good :( 

https://prnt.sc/TdtTqL7TrsaI

MaxSchmide:
> First screen does not look good :(
> 
> https://prnt.sc/TdtTqL7TrsaI

Yeah, I know :(. That's why had container before. But I don't know how to make it right using background properties. Tried change size and place but not found solution

mgrinko (changes requested):
Перевір себе по чеклісту будьласка

src/styles/layout/_main.scss
```diff
@@ -0,0 +1,19 @@
...
+  grid-template-rows: auto;
+  row-gap: 164px;
+  @include onLaptop {
+    padding: 120px 72px;
+    row-gap: 120px;
+  }
+
+  @include onMobile {
+    padding: 60px 20px;
+    row-gap: 90px;
```
mgrinko, line 17: But what if we need different distance for different sections?

mgrinko (changes requested):

meforoma (changes requested):
almost there 👍🏼 
few polishing moves remain.

- we don't want phone to fade-in on mobile layout

Monosnap MyBike 2023-06-05 17-13-04

---

- let's fix the text-area font to match the rest of the form

text-area font

meforoma (approved):
👍🏼

PR #20, opened 2023-05-29
finished task

artkas03:
- Deploy and create a Pull Request with a [DEMO LINK](https://artkas03.github.io/layout_landing-page/)


mgrinko (changes requested):
- Backgrounds should not be limited on large screens
- Bike should be centered

mgrinko (changes requested):
I'd google by the bike image to find one with higher resolution
Because the limited image looks really strange on a large screen

mgrinko (changes requested):
I see that header BG is still limited
because the header is inside container

mgrinko (changes requested):
Перевір себе по чеклісту будьласка

anstsot (approved):
Great job, approving:sparkles:

PR #42, opened 2023-06-01
add task solution

vasylynaBC:
[DEMO LINK](https://vasylynaBC.github.io/layout_landing/)


src/styles/block/contact.scss
```diff
@@ -0,0 +1,113 @@
...
+    @include OnTablet {
+      grid-column: span 3;
+    }
+
+    @include OnDesktop {
+      grid-column: span 5;
+    }
+
+    &--input {
+      box-sizing: border-box;
```
VitaliyBodnar, line 32: box-sizing: border-box should be applied to every element
* { 
 box-sizing: border-box;
}

src/styles/block/nav.scss
```diff
@@ -0,0 +1,36 @@
...
+    margin: 0;
+    padding: 0;
+    list-style: none;
+    display: flex;
+    flex-direction: column;
+    gap: 32px;
+  }
+
+  &__link {
+    color: #{$color-text};
```
VitaliyBodnar, line 12: try to use the same format for all variables
```suggestion
    color: $color-text;
```

VitaliyBodnar (changes requested):
Demo link is not valid

src/styles/block/contact.scss
vasylynaBC: If add to all doc, header is wrong position

vasylynaBC: image

etojeDenys (changes requested):
1. Use logo as a favicon
2. Change text color on hover for phone, email and address
3. Page shouldn't be reloaded on form submit, but the fields should be cleared. Use:

SanyaBratashchuk (changes requested):
1. Add favicon
2. Add transition for hover effects 
3. Fix textarea(placeholder)
image

loralevitska (changes requested):
You did an amazing job! Looks good! Let's do some improvements:

- make this number clickable 
image
- fix icons overlaping (don't show burger-menu icon when menu is open)
image
- disable scroll while menu is open
image
- fix image overlap. You can wrap each image in div and set property `overflow: hidden` for div and image will be increase like previous by overlap will be hidden
image
- add cursor pointer to buttons
- remove outline and default autocomplete styles 
image

loralevitska (changes requested):
Almost perfect)
You forgot:
- cursor pointer for clickable elements 
- and overlap not fixed for all parties (each image should scale within the size of its block)
image

vasylynaBC:
I cann't add overlap. It not work, need add fixed width for image.

VitaliyBodnar (approved):
Well done 🔥

PR #46, opened 2023-06-02
first commit

Faridun92:


etojeDenys (changes requested):
your demo link does not work

Volodymyr-Melnyk-Fron10der (changes requested):
1. this picture should be at the bottom
image

2. all pictures should be animated on hover

3. check paddings 
image

4. logo leads to the error 
image

5. check padding 
image
image

DarkMistyRoom (changes requested):
- check the top indent in the menu and in the header - they should be the same
image
image

- the logo and the "HOME" link in the menu lead to 404. fix it
- there should be a textarea in the form
image

- add scroll to the menu to handle the case when the screen is smaller than the menu block
image

sTorba24 (approved):
Good work!

PR #50, opened 2023-06-03
Develop

GGsmou:
[DEMO LINK](https://ggsmou.github.io/layout_landing-page/)


meforoma (changes requested):
looks good, few things to address

- i believe phone number should not fade-in on mobile layout

phone on mobile

---
- i see no required form fields alert

not required

---
- let's make this button navigate us to some section of the page

navigate somewehre

meforoma (approved):
love it 👍🏼 

the only thing i'd adjust is the tablet layout images size
they occupy the whole screen - and even more )

limiting the max-height sounds legit

image

PR #58, opened 2023-06-04
landing-page

Pashahu1:
https://pashahu1.github.io/layout_landing-page/


Volodymyr-Melnyk-Fron10der (changes requested):
1. add favicon 
image

2. 
image

3. check label's font family, phone number, address and email should be links
image

4. all images in gallery should have hover effect

5. phone number should appear on the icon hover
image

6. there is no text, the section's height should be 100vh
image

7. remove scrolling when the menu is open
image

8. check font family

volodymyr-soltys97 (changes requested):
This part is not made according to our design. Please check it
image
image

vlad-demchuk (changes requested):
1. form submission leads to error, it should just clear fields (proofread your code with the checklist)
image

2. these should be links (tel, mailto and so on)
image
image
image

3. this hover moves other blocks
image

4. header should be 100vh
image

5. transition works only for one side (check all such cases)
image

nazarmatsevych (changes requested):
1. Add scroll-behavior: smooth [rule](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior)

2. Make this links  anchors to appropriate section and don't open a new tab
image

3. Make sure to add transition style under general selector, not the one with :hover - this way transition will work smoothly both ways.

4. This links are not working (first link should make a call, second should open a mailer and third should open an address in google maps), same about phone icon in header. Remember than external links should open in a new tab
image
image

5. It's better to remove default form field value
image

6. Form should not reloads on submit, also you need to clear form fields on button click
image

7. Try to fix all of the previous comments before making re-request. If you have any questions - feel free to ask them in fe_chat

volodymyr-soltys97 (changes requested):
Here the text is not located as it is on the design
image
image
Check the fonts here to match the design
image
No need to open it in a new tab
image

Pashahu1:
I'm sorry for not being so attentive to this task and constantly asking for help for which I am very grateful, in the future I will try to correct errors before the pull request and conduct a thorough analysis of the task and the work itself

nazarmatsevych (changes requested):
1. Please add scroll-behavior: smooth [rule](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior) to make smooth scroll animation

2. It's better to hide menu on logo click
image

3. Logo should scroll to top of the page
image

dist/index.html
```diff
@@ -0,0 +1,226 @@
...
+      <div class="header__content">
+
+        <div class="top-bar">
+          <a href="#Recommended" class="top-bar__link-page">
+            <img src="/bose.580ac75f.svg" alt="Woman" class="top-bar__page-logo">
+          </a>
+        </div>
+
+        <div class="top-bar__panel" id="+1 234 555-55-55">
+          <a href="tel:+1 234 5555-55-55" class="top-bar__phone">
```
anastasiiavorobiova, line 25: Remove dashes and spaces from the href attribute for the phone number to avoid errors while redirecting

anastasiiavorobiova (changes requested):
Nice looking!
There are some tiny details to pay attention to: 
Screenshot 2023-06-10 at 15 29 19
The line on hover is too wide

Screenshot 2023-06-10 at 15 30 18
I don't have the possibility to scroll the menu in landscape mode on a mobile phone

Screenshot 2023-06-10 at 15 31 00
It's better to apply transformations directly to images, the whole section looks messy on hover.

dist/index.html
Pashahu1: tried to make it as good as possible

nazarmatsevych (changes requested):
Hey, almost done, please move scroll-behavior rule to html selector and check it out

Pashahu1:
okay i did it

nazarmatsevych (approved):

PR #69, opened 2023-06-05
add task solution

IraShtynda:
- [DEMO LINK](https://irashtynda.github.io/layout_landing-page/)


meforoma (changes requested):
great start 💪🏼 
few things to polish

- in mobile layout the footer acts weird - you can scroll past the footer

image

---

- and in the header we get the whitespaces

image

---

- not entirely sure, but feels like phone numbers should show on hover in tablet and desktop layouts

image

PR #72, opened 2023-06-06
@Dimoshua Landing-page Nothing v.2

Dimoshua:


loralevitska (changes requested):

- add phone number by icon hover
image
- avoid element jumping on page (burger menu icon by hovering). Fix all cases if present
- image hovering is overlap another pictures. You can wrap each image in div and give it property `overflow:hidden` so images will be increase but in their boards
image
- fix cursor in text-area it should be in the start of field
image
- fix textarea size 
image
- open address link in a new tab
- fix menu on mobile mode (and burger menu icon) according design
image

PR #73, opened 2023-06-06
Develop

Shpurka:
- [DEMO LINK](https://Shpurka.github.io/layout_landing/)


polosanya (changes requested):
Demo link doesn't work(
Make sure that you deployed everything

PR #85, opened 2023-06-07
add task solution

irinabolbat:
[DEMO LINK](https://irinabolbat.github.io/layout_landing/)


tetiana-ivanchenko (changes requested):
GJ, just small fix needed:
 buttons hover effects should have transition (actually all hover effect should)

tetiana-ivanchenko (approved):

PR #95, opened 2023-06-07
Landing-page Nothing v.3

Dimoshua:


DarkMistyRoom (changes requested):
- the logo in the header and in the menu should be in the same place

- the hover effect should work smoothly in both directions
image

- export images in higher resolution to make them look good on the scale
image

- the content should be centered
image

- make the hover effect for the button smooth too
image

- it should be a link to the google maps
image

- there should be a placeholder in the textarea too
image

- check the menu width on the tab and mobile
image

- 
image

PR #99, opened 2023-06-07
Add landing page

NataliGru:
- [DEMO LINK](https://nataligru.github.io/layout_landing-page/#)


tetiana-ivanchenko (changes requested):
1. hover effects shouldn't overlap the others elements
image
2. All form fields should be required
3.Increase all images on hover
4. it shouldn't be horisontal scroll on all devices

And678 (approved):
Good work, this is finished :heavy_check_mark:

PR #102, opened 2023-06-07
add landing-page

AnastasiiaKhemych:
[DEMO LINK](https://anastasiiakhemych.github.io/layout_landing-page/)


tetiana-ivanchenko (changes requested):
1. all clickable elements shoul have cursor pointer
2. all hover effects should have transition
3. Change text color on hover for phone, email and address

PR #104, opened 2023-06-07
first layout

ivanbilotserkivskiy:
https://ivanbilotserkivskiy.github.io/layout_landing/


tetiana-ivanchenko (changes requested):
1. increase each photo on hover instead of increasing whole block
image
2. hover effects shouldn't overlap the others elements
image
3. here should be the links
image
4. very poor quality
image
5. here should be the links too
image
image
6. Ensure that all nav__links work smoothly
7. Change text color on hover for phone, email and address


actually, before reauest review you should make shure that all point from the chacklist are completed

src/index.html
```diff
@@ -1,13 +1,381 @@
...
+                  required
+                >
+
+                <input
+                  type="text"
+                  placeholder="MESSAGE"
+                  class="feedback__message"
+                  required
+                >
+                <button class="feedback__send">Send</button>
```
juljaVysotska, line 370: Cursor doesn't change appearance when I hover on button.

juljaVysotska (changes requested):
I found some issues, please check and fix it.

And678 (approved):
All fixed now, good work!

PR #108, opened 2023-06-07
completed page

ovynarchuk:
[DEMO LINK](https://ovynarchuk.github.io/layout_landing-page/)


tetiana-ivanchenko (changes requested):
1. The speed of animations should be the same throughout the page (for example, increasing when hovering or moving blocks when scrolling)
2. All form fields should be required and have correct types (email, tel, etc.)
3. Page shouldn't be reloaded on form submit, but the fields should be cleared.
4. Hover effects shouldn't overlap the others elements
image
5. use svg format
image

YegorKorenkov (changes requested):
Good job! Please check comments below:

1. Element shouldn't overlaps another on hover, please check each case
image

2. Please, check fonts, they're different
image

3. Try to compress image, 1.3MB it's a lot
image

4. The same, 6MB too much, especially for users with bad internet
image

5.

And678 (commented):
Good work, approve :heavy_check_mark:

PR #109, opened 2023-06-07
add landing

AnatoliiLykhohodenko:
[DEMO LINK](https://AnatoliiLykhohodenko.github.io/layout_landing/)


tetiana-ivanchenko (changes requested):
1. Page shouldn't be reloaded on form submit, also shouldn't change the link
2. fix your layout fot the mobile and the tablet views

YegorKorenkov (changes requested):
Nice work 🔥please check minor miskateks below
1. Different fonts, please fix
image

2. The image is too large, 6MB is a lot, especially for users with poor internet, try reducing the size
image

3. Bg image looks different in design. Try repositioning your picture so you can see the steering wheel and front wheel
image

anstsot (approved):

PR #113, opened 2023-06-07
add task solution

olena-ole:
[DEMO](https://olena-ole.github.io/layout_landing-page/)


And678:
Message should be placeholder, not the text in textarea
image

And678:
Full width image is very low quality, there is a better version of this image in the mockup
image

olena-ole:
Thanks for the feedback, @And678 !
A small question about the image: would you recommend svg or just 2x quality for the header background? I used the biggest one, but 1x is not enough, probably(

vladmakohon:
> Thanks for the feedback, @And678 ! A small question about the image: would you recommend svg or just 2x quality for the header background? I used the biggest one, but 1x is not enough, probably(

I guess the problem is that you downloaded it from the tablet mockup instead of the desktop one.

tetiana-ivanchenko (changes requested):
1. fix according to checklist
image
2. hover effects shouldn't overlap the others elements
image
3. Increase all images on hover 
image
4. add some indentatons
image
5. Make sure everything looks neat on tablet and without horizontal scrolling

olena-ole:
@tetiana-ivanchenko thanks for the feedback!
Regarding the first suggestion (fix according to checklist), what exactly do you mean? The only idea I have is 'All addresses should be links to some location on Google Maps. Open then in a new tab with target="_blank"', but in the snippet there are only titles and working hours of the museums. Maybe I should add address to the title?

anstsot (approved):

PR #114, opened 2023-06-08
add task solution

klmyana:


tetiana-ivanchenko (changes requested):
1. hover effects shouldn't overlap the others elements
2. Placeholders should give exaples of what should be added, to make the expected format clear
3. text shouldn't starts here in the textarea
image

tetiana-ivanchenko (approved):
GJ 👍

PR #119, opened 2023-06-08
task done

tecHamster:
-[DEMO LINK](https://tecHamster.github.io/layout_landing/)


qoa11a (changes requested):
- When the phone icon is being hovered the mobile phone number should be shown.
- "Browse Nothing products by category" should be in 2 rows.
- "View all" button shouldn't be full width.
- Add validation to the form inputs. Make fields required.

polosanya (changes requested):
Almost perfect!
1. Let's add hover effects for all images
image
2. Pay attention that the price should be in one line 
image
3. The gap between links is too big
image
image

polosanya (approved):
Well done! 
Please make sure that you have the correct background color
image

PR #124, opened 2023-06-08
added task solution

tetianakostiuk1:
[DEMO LINK](https://tetianakostiuk1.github.io/layout_landing-page/)

### DIFF
--- a/src/styles/blocks/contacts.scss
+++ b/src/styles/blocks/contacts.scss
@@ -80,7 +80,14 @@
     font-weight: 400;
     font-size: 14px;
     line-height: 140%;
+  }
 
+  &__phone,
+  &__email,
+  &__address{
+    &:hover {
+      color: #818181;
+    }
   }
 
   &__href {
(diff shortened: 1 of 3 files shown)


tetiana-ivanchenko (changes requested):
fix all points from the checklist (points: 1, 6, 10 (doesn't works at all), 11, 15)
also message should be textarea

YegorKorenkov (changes requested):
1. Logo should be dark in menu
image

2. Try reducing the image size, because 6 MB is a lot, especially for people with poor internet. To do this, for starters, you can use a jpg image format instead of png and then use some compressing services
image

3. Please remove target="_blank" for phone, redundant
image

4. The same
image

5. Please check position of your bg image, should be like in design
Your landing:
image


Design:
image

6. Title should be on one line
image

7. Too big logo for mobile screens, please check the right size in the design
image

tetiana-ivanchenko (changes requested):
Change text color on hover for phone, email and address

anstsot (approved):

PR #129, opened 2023-06-08
add landing MET 

Medaed:
[DEMO LINK](https://medaed.github.io/layout_landing-page/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -226,7 +226,11 @@
                 </div>
               </article>
             </div>
-            <button class="now-on-view__button" type="button">View all exhibitions</button>
+            <button
+            class="now-on-view__button"
+            type="button"
+            onclick="window.open('https://www.metmuseum.org','_blank')"
+            >View all exhibitions</button>
           </section>
 
           <section class="explore-online" id="explore-online">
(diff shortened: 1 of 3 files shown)


loralevitska (changes requested):

- add phone numbers by hover on phone icon
image
- add cursor pointer to each clickable elements
- fix image overlap (you can wrap each image to div and set `overflow: hidden` property)
image
- fix page background color and form colors
image
- remove outline and autocomplete default properties
image
- add cursor pointer and hover effect to clickable elements
image
- fix this link
image
- fix images on mobile mode 
image
- disable textarea autosize
image

SanyaBratashchuk (changes requested):
1. Disable textarea resize 
2. Add hover effects for these links 
image
3. Fix mobile layout 
image

TarasHoliuk:
1. Fix colors - the page background should be white:

The App:
image

Mockup:
image

2. Don't add a border on input focus - it moves all elements located below each time. Change outline styles and use it OR make border-color transparent by default.

3. Textarea size shouldn't be editable (fixed size which is required by design)

4. The footer in the mobile version looks not good:
image

5. The right image in the mobile version looks not good (change positioning):
image

TarasHoliuk (changes requested):
Well done :+1: But a few things still might be improved (check comments)

VitaliyBodnar (changes requested):
Well done 🔥
Let's improve your solution 🧑‍💻

the form fields are transparent and not as in the design
Screenshot 2023-06-11 at 17 34 14

Medaed:
What the color it shoud be ? I have same color as on design 
Screenshot 2023-06-11 at 18 10 40

DarkMistyRoom (changes requested):
You're almost there!
- I can't reach the phone numbers. When I try to do it, they disappear
image

- fix the menu for the small screens. remove the white indent at the bottom and fix the situation when top actions stay on the top while other content scrolling
image

loralevitska (changes requested):
Almost perfect) Just a few moments should be fixed:

- fix form indent for form according design
image
image
- would be better if something happens by click on this button. Try to find original MET site and navigate to their link. It make your site more interesting)
image

SanyaBratashchuk (approved):

PR #145, opened 2023-06-09
Develop landing-page

taras-bilas:
[DEMO LINK](https://taras-bilas.github.io/layout_landing/)


YegorKorenkov (changes requested):
1. Try reducing the image size, because 7.5 MB is a lot, especially for people with poor internet. To do this, for starters, you can use a jpg image format instead of png and then use some compressing services
image

2. Disable page scrolling under the menu
image

3. Please add hover effects to images
image
image

4. The same
image

anstsot (changes requested):
Good start:four_leaf_clover:
Let’s add a few fixes to improve your task.
1. try to set the right position for the image
image

2. add hover effects for all links and needed elements
image

3. add a smooth transition for hover
4. fix the font for textarea in the form
image

5. center the blocks
image

YegorKorenkov (approved):
Nice work! Approve, but please check my comment and make changes

1. Please add more space between submit button and phone link
image

PR #146, opened 2023-06-10
MyBIKE

RomanPerepelkin:
[DEMO LINK](https://RomanPerepelkin.github.io/layout_landing-page/)


nazarmatsevych (changes requested):
1. Make sure to add transition style under general selector, not the one with :hover - this way transition will work smoothly both ways.

2. Images should not overlap each other
image

3. Add curson: pointer and hover effects to all buttons 
image


4. Remove this weird hover effects from form fields
image

5. First link with mobile number should make a call and [second should open a mailer](https://css-tricks.com/snippets/html/mailto-links/)
image
image

nazarmatsevych (approved):
Well done!

1. It's better change or remove this hover effect
image

PR #147, opened 2023-06-10
add task solution

kylyvnyk:
DEMO LINK - https://kylyvnyk.github.io/layout_landing-page/


YegorKorenkov (changes requested):
1. Try reducing the image size, because 6 MB is a lot, especially for people with poor internet. To do this, for starters, you can use a jpg image format instead of png and then use some compressing services
image

2. Please check this block according to design

Your landing:
image

Design:
image

3. hover image should not ovrlaps
image

4. Missed placeholder
image

5. Please check button size accroding to design
image

6. Title should be centered
image

src/index.html
```diff
@@ -1,13 +1,271 @@
...
+        <a href="#" class="top-bar__logo-link">
+          <img
+            src="styles/images/logo-whitex2.png"
+            alt="mybike logo"
+            class="top-bar__logo"
+          >
+        </a>
+
+        <div class="top-bar__icons">
+          <a
```
juljaVysotska, line 30: Try to use `svg` for icons for better quality in future.
image

juljaVysotska (approved):
Good job!

PR #156, opened 2023-06-11
add task solution

purpleFade:
https://purplefade.github.io/layout_landing-page/


loralevitska (changes requested):
Almost perfect) Let's improve

- remove outline and autocomplete default styles. Read more about [changing autocomplete styles](https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/)
image
- fix 404 error for address link

PR #168, opened 2023-06-13
nothing landing

Livan94:
- [DEMO LINK](https://livan94.github.io/layout_landing-page/)


Andriy-Shymkiv (changes requested):
good job! need to improve some things: 
- add margin after images here:
image
- add `transiton` for your buttons on hover to make it more smoother
- better to set `margin-bottom` for img and not to give `margin-top` for title in this case:
image please change it everywhere

PR #169, opened 2023-06-13
add task solution

Klishchar-Ihor:


vlad-demchuk (changes requested):
demo link isn't working

vlad-demchuk (changes requested):
1. checklist
image

2. also, number should be shown on hover according to design
image
image

3. 
image

4. these should be links
image

5. form field should be required
image

6. 
image
image

7. center the form between mobile and tables widths
image

PR #180, opened 2023-06-15
add task solution

MariaSnegireva:
[DEMO LINK](https://MariaSnegireva.github.io/layout_landing/)

[Figma Design] (https://www.figma.com/file/lSR1m42L9YwzQwzzxKwHpw/THE-MET?type=design&node-id=8590-537&t=Mb11dFaUUH48PuET-0)


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
When I want to select a phone number with the mouse, this window disappears, it is necessary to provide for this possibility
image
Add transition for these elements when on hover
image

src/styles/blocks/contacts.scss
```diff
@@ -0,0 +1,112 @@
...
+
+    &-block {
+      display: flex;
+      flex-direction: column;
+      row-gap: 48px;
+
+      @include onDesktop {
+        @include pageGrid;
+        grid-column: span 6;
+
```
volodymyr-soltys97, line 57:
```suggestion
```

PR #195, opened 2023-06-19
layout mybike add all block

petrushinsky-kirill:
[DEMO LINK](https://petrushinsky-kirill.github.io/layout_landing-page/)


Anastasiia-Svintsova (changes requested):
- Check the text width on the desktop 
image
image

- Add hover effects to the links 
image

- The user shouldn't be able to submit the form if there is no message
image

- This text should always be in one line according to the design 
image

- The footer has 0 height on the mobile. Fix it, please 
image

Anastasiia-Svintsova (changes requested):
- Hey! I still can't see the footer image on the mobile 
image

- Fix the logo links, please. There should be just an anchor link to the top of the page. The page shouldn't reload on the logo click
image

Anastasiia-Svintsova (approved):

PR #199, opened 2023-06-19
Develop

RUBIC0N5545:
https://RUBIC0N5545.github.io/layout_landing-page/


etojeDenys (changes requested):
1. Use logo as a favicon
2. All addresses should be links to some location on Google Maps. Open then in a new tab with target="_blank".
3. Disable page scrolling under the menu 
4. Page shouldn't be reloaded on form submit, but the fields should be cleared
5.  also check the layout and fix all mismatches

VitaliyBondarenko1982 (changes requested):
Good job.
Needs some improvements.
- not fixed prev comment about disable page scrolling with open menu
Screenshot from 2023-06-22 13-58-33
- add transition property for all elements which have hover effect.
for example on this elements no transition:
chrome-capture-2023-5-22
- message field should be textarea. also consider to make custom scrollbar for textarea
Screenshot from 2023-06-22 13-52-17
https://www.w3schools.com/howto/howto_css_custom_scrollbar.asp
- fix all mismatches.
  for example here mismatch text and font-size.
  design:
Screenshot from 2023-06-22 13-57-25
actual:
Screenshot from 2023-06-22 13-57-13
check and compare with design other places
- add cursor pointer for send button. check other interactive (buttons, links) elements.

etojeDenys (changes requested):
1. we don't need so much free space on mobile
image
2. add smooth scroll
image

RUBIC0N5545:
> 1. we don't need so much free space on mobile
> 
> image 2. add smooth scroll image

I don't understand what's wrong and how it must look correctly

VitaliyBondarenko1982 (approved):
Looks good to me.

PR #213, opened 2023-06-21
Landing-page Nothing v.7

Dimoshua:


DarkMistyRoom (changes requested):
Good job! Few moments:
- the number should be reachable. But when I try to hover over it, it disapears
image

- set max width to the container to 1260px to limit the width of the content according to the mockup. max-width should be hardcoded but side margins - flexible
image

- fix this button on the mobile
image

PR #215, opened 2023-06-21
header and menu mobile

Elijah900BCE:
[DEMO LINK](https://elijah900bce.github.io/layout_landing-page/)


agniya-i (changes requested):
Keep working 👍

1. Check the position of the background image and the color of logo, it doesn't match the design 
2. Add a hover state for menu elements

volodymyr-soltys97 (changes requested):
To improve:
1. Add favicon to page
image

2. Need to fix background-image. You also need to add indents to the right and left. Check on design
image

3. Need to show phone and call here
image

PR #218, opened 2023-06-22
add landing-page

TVChulkova:
[DEMO LINK](https://TVChulkova.github.io/layout_landing/)


volodymyr-soltys97 (changes requested):
Good start 👍 
Let's to fix some mistakes.
1. You need to add cursor: pointer for all clickable elements
image

2. You also need to add transition that the menu appears smoothly and you also need to prohibit scrolling when the menu is open so that there are no such situations as on the screen
image
image

3. Add transition for all clickable elements, for example buttons

4. Make these fields is required before sending
image

src/index.html
anastasiiavorobiova, line 1: validate your HTML with https://validator.w3.org/nu/#textarea there are a lot of things to fix

anastasiiavorobiova (changes requested):
Great work! Keep doing: 
Screenshot 2023-06-23 at 15 05 05
Unable to click on the appeared phone numbers
Screenshot 2023-06-23 at 15 05 11
Add transitions and cursor pointer for all clickable elements
Screenshot 2023-06-23 at 15 07 32
Screenshot 2023-06-23 at 15 08 45
Check styles for forms, it differs from the design including the font-family property
Screenshot 2023-06-23 at 15 10 03
Check indentations
Screenshot 2023-06-23 at 15 09 39
it should be scrollable for mobiles in landscape mode, consider adding overflow y property

etojeDenys (changes requested):
1. let's use textarea instead the input tag for this field to have ability write in several lines(but disable the scaling)
image



2. it also should be links with some hover
image


3. desktop layout looks broken
image
image

volodymyr-soltys97 (changes requested):
1. Need to show the phone numbers when hovering this icon
image

2. The user will not be able to change the size of this field
image

image

4. Need to remove the autocomplete styles for inputs
image

5. Add font weight for this button as on the design
image

VitaliyBondarenko1982 (approved):
Looks good to me.

PR #220, opened 2023-06-22
Landing-page Nothing v.8

Dimoshua:


loralevitska (changes requested):
Almost perfect! Please pay attention to my comments:

- according to the design, when hovering over the phone icon, the icon itself should remain, and only the phone should appear
your version
image
design version
image
- add cursor pointer to all clickable elements 
- remove outline, it breaks the design
image
- the burger menu icon and the cross icon in the menu should be in the same place, fix it. I think the reason may be this indentation in the cross icon
image
- don't show phone number on mobile mode. It breaks the design
image
- remove default autocomplete background-color, it breaks the design
image

marina-tilniak (changes requested):
test reject

PR #226, opened 2023-06-24
B&O Landing page code

oshapkun:
https://oshapkun.github.io/layout_landing-page/
Design https://www.figma.com/file/DtkQmQ797hk0nI4KfMi2Uq/BOSE-New-Version?type=design&node-id=6817%3A268&mode=design&t=XZRqprZkZsN4D58H-1


Anastasiia-Svintsova (changes requested):
Great start! Let's add some improvements 

- Add a correct hover effect to the phone logo. There should be a phone number as well 
image

- The message field should be required as well 
image

- Fix the image size. It doesn't match the design for now  
image
image

volodymyr-soltys97 (changes requested):
You need to add transition for all buttons when hover.

volodymyr-soltys97 (changes requested):
Add `cursor: pointer` for this button
image

Also add transition for these elements when on hover
image

No need to show your phone on the mobile version
image

src/index.html
```diff
@@ -1,13 +1,313 @@
...
+    <header class="header">
+      <div class="container">
+        <div class="header__content">
+          <div class="top-bar">
+            <a href="#" class="top-bar__logo-link">
+              <img
+              src="images/bo-logo-2x.png"
+              alt="b&o logo"
+              class="top-bar__logo"
+              >
```
volodymyr-soltys97, line 25: Add 2 spaces for attributes and check it everywhere

src/index.html
```diff
@@ -1,13 +1,313 @@
...
+            </a>
+            <div class="top-bar__icons">
+              <a href="tel:+1 234 5555-55-55" class="icon icon--phone"></a>
```
volodymyr-soltys97, line 28: Need to fix format number in href everywhere
```suggestion
              <a href="tel:+123455555555" class="icon icon--phone"></a>
```

volodymyr-soltys97 (changes requested):
You need to check why horizontal scroll appears on the mobile version
image

mykhalenych (approved):

PR #228, opened 2023-06-24
add task solution

atrofymovych:
[DEMO LINK](https://atrofymovych.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good start 👍 
Let's to fix
1. Add favicon to page
image

2. Need to fix paddings and background-image
Demo:
image

Design
image

3. You need to show phone when hovering this icon
image

4. Add transition when this menu appear on page
image

PR #234, opened 2023-06-26
Nothing(R)

levchenko-dmytro:


volodymyr-soltys97 (changes requested):
Good start 👍 
Let's to fix some moments
1. Add title on page
image

2. Add `cursor: pointer` for this button
image

3. Remove default styles after autocomplete
image

4. Add additional styles for these elements when on hover
image

levchenko-dmytro:
The name of the title is "Nothing(R)"
image

loralevitska (approved):
Good job! Please improve one more point

PR #238, opened 2023-06-26
the met

VladyslavKolisnyk:
DEMO LINK https://vladyslavkolisnyk.github.io/layout_landing-page


volodymyr-soltys97 (changes requested):
Good work 👍 
Let's to fix some moments
Add transition and cursor: pointer for all clickable elements when on hover
image
image

Need to remove default styles after autocomplete
image

loralevitska (changes requested):

- still don't fixed 
image
- disable text-area resize, it breaks design
image
- also add cursor pointer to all interactive elements

volodymyr-soltys97 (changes requested):
Check indents on mobile version
image

Add additional styles for these elements when hovering
image

src/index.html
anastasiiavorobiova, line 1: Check the validity of your HTML with [HTML validator](https://validator.w3.org/nu/#textarea)

anastasiiavorobiova (changes requested):
Really nice looking landing! Let's make it perfect

anastasiiavorobiova (approved):
Awesome!

PR #250, opened 2023-06-30
add header + aside

VvynnykV:
[DEMO LINK] (https://VvynnykV.github.io/layout_landing-page/)

**_### Please check if I correctly used the following tricks in my project:_**

Disable page scrolling under the menu using the next code:
}

Make it possible to scroll the menu if it is higher than the viewport;
.menu {
}
.menu__top {
  top: 0;
  z-index: 1;
  background-color: #e5e5e5;
}

The problem that I noticed that ".icon--close" (the cross) its change his position on horizontal line if I add the code that dont make it possible to scroll


VitaliyBondarenko1982 (changes requested):
Good start.
Disable scroll and scroll open menu looks good.
Of course without scrollbar cross-icon visually move to right, but it is expected behavior
Waiting other sections

VvynnykV:
thanks ro review. what do you mean "Waiting other sections"? I think this task has done and all another sections its for future tasks

VitaliyBondarenko1982:
> thanks ro review. what do you mean "Waiting other sections"? I think this task has done and all another sections its for future tasks

now demo links looks such. and you will be implement other landing sections from design. we are waitning them
and in this case better to ask in fe_chat and do not make re-request just for it)
chrome-capture-2023-5-30

VitaliyBondarenko1982 (changes requested):
.

PR #270, opened 2023-07-03
NOTHING landing

Horizon-git:
[DEMO LINK](https://Horizon-git.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good start 👍 
Let's to improve your page
1. You need to make a call when clicking on the phone
image

2. Need to remove default styles for input after autocomplete
image

3. Need to remove default border for this button
image

4. When I move the cursor here, these fields get a little bigger and jump around on the page, I need to fix that
image

VitaliyBondarenko1982 (approved):
Looks good to me

PR #282, opened 2023-07-05
product

BOHDANLIUYUI:
[DEMO LINK](https://BOHDANLIUYUI.github.io/layout_landing-page/)

### DIFF
--- a/src/styles/blocks/_product.scss
+++ b/src/styles/blocks/_product.scss
@@ -2,7 +2,7 @@
   transition: transform 0.3s;
 
   &:hover {
-    transform: scale(1.2);
+    transform: scale(0.9);
   }
 
   &__photo {


etojeDenys (changes requested):
1. should be clickable and add cursor pointer
image


2. Page shouldn't be reloaded on form submit, but the fields should be cleared.

4. should be clickable
image


5. Disable page scrolling under the menu 
image



6. Increase all images on hover (make them links to #, if there is no better option)

IvanFesenko (changes requested):
Design mismatch. Section should have 100%vh. Icons have background - export just phone and burger icon without background

https://github.com/mate-academy/layout_landing-page/assets/61904995/2c28dc14-4acf-45cb-a81c-a7876a21bc96

2023-07-06_15h17_09

2023-07-06_15h21_27


All fileds in form should be required

Add scale effect to this images on hover as well
2023-07-06_15h20_29

Viktor-Kost (changes requested):
Hello!)

### To improve:
1) not header__content should be 100vh, but the full header. And then - just position title correctly
https://github.com/mate-academy/layout_landing-page/assets/117587463/63de4d48-3512-4b23-b04b-bc510a1c0e72

2) logo link should have `href` and `cursor:pointer`
3) bigger distance than needed
Screenshot 2023-07-06 at 21 02 28

4) you should use three different background images for mobile, tablet and desktop. Take these images from each layout at figma and change it using media queries

etojeDenys (changes requested):
your link does not work, try to re-deploy

etojeDenys (approved):
GJ!

PR #290, opened 2023-07-08
add Bike landing

natalia-serogina:
[DEMO LINK](https://natalia-serogina.github.io/layout_landing-page/)
https://natalia-serogina.github.io/bike-landing/


sTorba24 (changes requested):
Good work! Lets improve it

1. I still see menu icon when i click and not unhover yet
image

2. Good effect, but would be good to use it only once

try to move scroll up and down fast and you'll see what i mean :)
image

3. decrease scaling, it shouldn't overlap other img
image


4. it should be arrow up 
image

polosanya (changes requested):
A few more fixes left:
1. Background color doesn't match the design
image
image
2. Download the image without button in right bottom corner
image
3. This button does nothing on click
image
4. Placeholders don't match the design
image
image

PR #291, opened 2023-07-08
Develop

HannaDevnina:
[DEMO LINK] https://HannaDevnina.github.io/layout_landing-page/
Hello.
I didn't understand exactly should I finish landing page right now or it will be next task. There are a lot issues  in the checklist which we haven't learned yet .


VitaliyBodnar (changes requested):
Answer to your question:
In this task, you don't need to finish the whole template. Need to finish only the header part. 

- the bad quality of the background image 
Screenshot 2023-07-08 at 14 38 09

- you need to show the phone number on hover to the phone icon 
Screenshot 2023-07-08 at 14 39 01

- there shouldn't be a scroll on the mobile menu
Screenshot 2023-07-08 at 14 41 20

HannaDevnina:
Hello. 
I've changed background image and remove scroll but I didn't get how to create pop-up phone number. I couldn't find any mention how it should be done in the video, must be I missed something. Also I thought it's could be some video had been changed. I have two option I decribed they in files in comments. Please, help me.

src/index.html
```diff
@@ -1,13 +1,138 @@
...
-  <body>
-    <h1>Landing Page</h1>
+  <!-- #endregion -->
+  <body class="page__body" id="page__body">
+    <header class="header">
+      <div class="top-bar">
+        <a href="#page__body" class="top-bar__logo">
+          <img src="images/logo.png" alt="The Met logo">
+        </a>
+        <!-- розглядала варіант створити блок який буде мати
```
TarasHoliuk, line 32: it will be the preferable case since that should be real links
ask in the chat if you need some help with the implementation (you should use combined (chained) selector)

TarasHoliuk: P.S. Just for the future - it's better to ask such questions in the chat - the response time will be much shorter.

TarasHoliuk:
1. Links in the menu should have `:hover` effect:
image

2. The menu should appear/disappear smoothly (If you able to do it now)

TarasHoliuk (changes requested):

PR #298, opened 2023-07-09
Develop

maxonshev:
https://maxonshev.github.io/layout_landing-page/


VitaliyBodnar (changes requested):
Well done 🔥
Let's improve your solution 🧑‍💻

- you need to show the phone number on the hover
Screenshot 2023-07-09 at 16 58 24

- wrong spaces
Screenshot 2023-07-09 at 16 59 23
Screenshot 2023-07-09 at 16 59 16

DarkMistyRoom (changes requested):
- icon and number should be `tel` links. And the number should be black according to the mockup
image

- set `font-weight` for these lines
image

- images in this section should be 100% width according to the mockup(tab screen)
image

- add some margins to make the menu looks good in landscape mode
image

SanyaBratashchuk (approved):

PR #311, opened 2023-07-11
landing page is ready

rimidbit:
[DEMO LINK](https://rimidbit.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good start 👍 
Let's improve your page
1. Need to check font-size for phone. Also add a transition when this phone appears on page
Demo
image

Design
image

2. You need to make field name required
image

3. Add additional styles for these elements when hovering
image

volodymyr-soltys97 (changes requested):

polosanya (changes requested):
A few fixes left
1. Icons don't match the design. Make sure to add them as images so the size is correct
image
image
2. Form doesn't match the design. Pay attention to the inputs
image
image
3. Justify the text correctly on tablets
image
image

polosanya (approved):
Great! Well done! 🔥

PR #322, opened 2023-07-13
Develop

dkornienko31:
[DEMO LINK](https://dkornienko31.github.io/layout_landing-page/)


polosanya (changes requested):
Nice start!
1. Phone link should be clickable with the ability to call
image
2. Make sure that you add a hover effect for the default state of the element, not hover, so it works both ways smoothly
hover
3. Check the indentation between text and price
image
image
4. Fix the button text
image
image
5. There should be no scroll in burger menu
image
6. Make sure that the images are centred. By the way, they should be bigger on tablets
image
image

TarasHoliuk:
[Checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md) 12 - add scroll **inside** the menu if it's needed

TarasHoliuk (changes requested):
Almost perfect :fire: 
Let's improve one more thing

polosanya (changes requested):
1. There is a typo in "View". Please fix it
image
image
2. Make sure that all interactive elements have cursor pointer on hover
image
3. Background color doesn't match the design
image
image
4. Comment about the scroll still not fixed. Make sure it's possible to scroll if the screen height is smaller than the content. Check yourself with DevTools. Users should be able to see all the content of the burger menu, but not scroll through the whole page like it was before 
image
image
Please ask questions in fe_chat if you have any difficulties

src/index.html
```diff
@@ -1,13 +1,309 @@
...
+                <a class="nav__link" href="#store">Store</a>
+              </li>
+              <li class="nav__item">
+                <a class="nav__link" href="#about us">About Us</a>
+              </li>
+              <li class="nav__item">
+                <a class="nav__link" href="#contact us">Contact Us</a>
+              </li>
+              <li>
+                <a href="tel:+1 234 5555-55-55" class="menu__phone-number">
```
volodymyr-soltys97, line 77: Need to fix format phone only in href everywhere
```suggestion
                <a href="tel:+123455555555" class="menu__phone-number">
```

volodymyr-soltys97 (changes requested):
Need to remove default styles for input after autocomplete
image

mykhalenych (approved):

PR #324, opened 2023-07-13
first try

IhorVerba:
[DEMO LINK](https://IhorVerba.github.io/layout_landing-page/)


polosanya (changes requested):
Nice start!
1. Phone number shouldn't disappear on hover. Also, it should be clickable with ability to call
phone
2. Pay attention to the font styles of price 
image
3. You can download this button without that button in the corner
image
4. There should be no error on form submit
image

volodymyr-soltys97 (changes requested):
1. Add `cursor: pointer` for this element
image

2. Need to fix background image on mobile
Demo
image

Design
image

Anastasiia-Svintsova (approved):

PR #325, opened 2023-07-13
solution 1.0

aliliolek:
[DEMO LINK](https://aliliolek.github.io/layout_landing-page/)


loralevitska (changes requested):
You did an amazing job! Looks good! Let's do some improvements (check comments)

- implement this text by phone icon hover according design
image
- add hover effect for this links
image
image
- fix this title on tablet mode according design, also add right line-height 
image

aliliolek:
1. Added a **block with the phone numbers** that appears on hover over the call-icon.
2. Added underline for **nav__link:hover**. (as on "call to order")
3. Fixed some issues:
- on **form__field:hover** border-color: #D5D7DE;
- on **form__field:focus** border-color: $main-dark;
However, I don't see in Figma design that **form field`s background-color** should change.

sTorba24 (changes requested):
Almost good! lets improve it

1. set just `href` attr and use tag `a` - link . But not onclick event
image

2.still not fixed autocomplete styles for input
image

PR #326, opened 2023-07-15
add head, header, aside and menu - the met

Iryna1802:
https://Iryna1802.github.io/layout_landing/


DarkMistyRoom (changes requested):
- please check the logo size. It looks too small
image

- the numbers should appear on the hover
image

- Please fix the menu according to the mockup
image

etojeDenys (changes requested):
fix prev comments then re-request me

polosanya (changes requested):
Good job! Keep working! 
Re-request the review when other sections are ready
1. Let's add a favicon for the page
image
2. Make sure that all interactive elements are clickable with cursor pointer. Also, user should be able to call by clicking on phone icon/number.
image

Iryna1802:
I fixed it. Link works. I am finalizing other sections.

volodymyr-soltys97 (changes requested):
1. Need to remove the scroll when this menu is open
image

2. Add cursor pointer and hover effects for all clickable elements on the all page

image

4. Need to remove the autocomplete styles
image

image

6. These elements should be links
image

sTorba24 (approved):
Well done!

remove scrolls on menu
image

PR #331, opened 2023-07-16
add task solution 1.0

mykyta01:
- [DEMO LINK](https://mykyta01.github.io/layout_landing-page/)


DarkMistyRoom (changes requested):
- check these indents. They look smaller in the mockup
image
image


- the logo in the menu is smaller than in the header
- please add another hover effect. There is no more blue color on the page - it looks a bit wierd
image

- make the hover effect smooth in both directions
image

- add a hover effect to all links
image

- check the height of this button
image

- don't reload the page on form submitting, just clear the form

- the menu close button and the open button should be in the same position, but they don't on the tab
- check all vertical indents, please

- looks weird. fix it
image
image

- make your background image looks the same as in the mockup
image

- add vertical scroll to the menu to make it works fine in the landscape mode
image

PR #337, opened 2023-07-17
Landing page

mykyta01:


src/index.html
```diff
@@ -1,13 +1,277 @@
...
+              Book a test ride
+            </a>
+          </ul>
+        </div>
+      </nav>
+    </header>
+
+    <main class="main">
+      <section class="page__section move">
+        <div class="container-1">
```
xiimiik, line 82: Why do we need all these different containers? "container-1", "container-2", etc.?
In order to set margins, we have a `page` block, with the help of which we position elements in it

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  @include onDesktop {
+    padding: 0 120px;
+  }
+}
+
+.footer {
+  z-index: -1
+}
```
xiimiik, line 53: Why are 3 blocks in one file?

src/styles/blocks/assortiment.scss
```diff
@@ -0,0 +1,113 @@
...
+  color: $text-white;
+
+  &__photo {
+    width: 100%;
+    height: 100%;
+    display: block;
+    transition: transform 0.5s ease;
+
+    &:hover {
+      transform: scale(120%);
```
xiimiik, line 11: opera_eXacEAUW1U

Scale overlaps neighboring elements, this should not happen

xiimiik (changes requested):
GJ, left few comments :)
Also, the menu animation works only when opening, you need to fix it.

mykyta01:
I have a problem: second commit don't deploy ( I wrote npm run deploy, wait < 10 min and nothing happened)

PR #338, opened 2023-07-17
Layout first landing

denedweb:
[DEMO LINK](https://denedweb.github.io/layout_landing-page/)


loralevitska (changes requested):

- this point needs to be corrected, it affects the understanding of the site as a whole, users and recruiters pay attention to this when viewing work
image
image
- disable scroll while menu is open and the content fits into the frame of the screen. 
image
so that the code works with one and another comment, add a scroll only when the content does not fit 
image
Try [overflow](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow)

VitaliyBodnar (changes requested):
- the pictures should not overlap each other on hover 
 Screenshot 2023-07-19 at 14 43 33

- need to add validation for the email 
Screenshot 2023-07-19 at 14 47 30

- the hover is too large, is supposed to be only on the text
Screenshot 2023-07-19 at 14 48 37

VitaliyBodnar (approved):
Good job!

PR #340, opened 2023-07-18
add landing layout

vladyslav-svorin:
- [DEMO LINK](https://vladyslav-svorin.github.io/layout_landing-page/)

i couldn't find a way to remove the vertical scroll using CSS


loralevitska (changes requested):
Keep up the good work! Some comments to fix:

- disable scroll while menu is open but leave if the content does not fit, use [overflow](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow)
image
- fix menu links, it should navigate to certain block
image
- fix 405 error after form submit
- make all form field required 
image
- phone and address in footer should be link
image
- fix page title 
image
- check yourself with design, according tablet design there shouldn't be 3 images
image
- 
image

loralevitska (changes requested):
Almost perfect) Let's improve

- these links doesn't work 
image
- address should navigate user to map (always open links in new tab)
image

etojeDenys (changes requested):
1. Page shouldn't be reloaded on form submit, but the fields should be cleared
2. Increase all images on hover (make them links to #, if there is no better option)

VitaliyBodnar (changes requested):
- these comments are still not fixed 
Screenshot 2023-07-21 at 14 36 00

etojeDenys (changes requested):
All form fields should be required and have correct types (email, tel, etc.)

loralevitska (changes requested):
A little more and you will be on target)
If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

- ass smooth navigate user from menu, now it look like teleport
- and this comment still not fixed 
image

etojeDenys (approved):
lgtm

PR #343, opened 2023-07-19
create landing header

AnastasiiaHombalevska:
- [DEMO LINK](https://AnastasiiaHombalevska.github.io/layout_landing-page/)


Anna-Tkachenko (changes requested):
Demo link doesn't work, check it please

AnastasiiaHombalevska:
i have problem with deploy :(

dffUqp (changes requested):
Good job! Some comments:

- Phone icon should be phone link (check point 5 in checklist https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md)
- Add some padding for your inputs and i see hardcode value to the textarea, use placeholder instead
- Check your Contact us section, you have only one column there.
- Lest remove ability to resize [textarea](https://developer.mozilla.org/en-US/docs/Web/CSS/resize)
- Would be better to use transition in place where you have hover
- Disable scroll when menu is open

roman-mirzoian (approved):
Well done 👍

PR #353, opened 2023-07-21
#1 Landing Page

julwer1k:
- [DEMO-LINK](https://julwer1k.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good start 👍 
Let's improve your page
1. Add `cursor: pointer` and transition for all clickable elements when hovering
image

2. Need to remove default styles for input after autocomplete and remove borders
image

vlad-demchuk (approved):
GJ

PR #365, opened 2023-07-24
My bike landing

Denys14Samoilenko:
[DEMO LINK](https://Denys14Samoilenko.github.io/layout_landing-page/)


BudnikOleksii (changes requested):
Looks great👍
Some moments that should be fixed in accordance to the checklist:
4. Change text color on hover for phone, email and address.
5. Make sure phone icons and phone numbers are real link to start a call.
7. Increase all images on hover (make them links to #, if there is no better option).

Anna-Tkachenko (approved):
Good job!
Just check the height for the textarea field according to the design, looks like it should be bigger

PR #379, opened 2023-07-26
add solution

AndrMar1939:
[DEMO LINK](https://AndrMar1939.github.io/layout_landing-page/)


src/styles/blocks/button-general.scss
```diff
@@ -0,0 +1,24 @@
...
+
+  font-family: inherit;
+  color: #fff;
+  font-size: 16px;
+  font-weight: 700;
+  line-height: 56px;
+
+  background-color: $secondary-page-color;
+
+  transition: all $transition-duration;
```
alexander-ignatow, line 14: try setting transition only to styles you need

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,36 @@
+.page {
+  scroll-behavior: smooth !important;
```
alexander-ignatow, line 2: Using `important` is a bad practice, unless you have a strong reason to do it.

src/styles/utils/normalize/normalize.scss
alexander-ignatow, line 1: This looks more like a reset :)

src/index.html
alexander-ignatow, line 1: Please re-check your layout on wide screens

image

alexander-ignatow (changes requested):
Good job!

I like your file structure 👍 

Please fix several minor issues

src/styles/blocks/page.scss
AndrMar1939: Yes, smooth doesn't work in my current version of Chrome without '!important'. I read Stackoverflow on this topic and found the following solution

src/index.html
AndrMar1939: I'm sorry, I didn't catch that. 
According to the layout, the maximum width is 1260px.
On wide screens, the wrapper is stretched to its maximum width and centered with margin: 0 auto

Anna-Tkachenko (changes requested):
Good job!
Some comments:
- check page displaying, I should not see any white indents before and after the header (at the right and left sides). Background should have 100% width
- will be great to show the contact number near the phone icon on hover state
- I can see other elements when menu is opened via `tab` button, check it please (https://prnt.sc/Tm2TQ3y8USBB)
- logo in the opened menu should also be the link

Moroz-Dmytro (changes requested):
1. Let's disable ability to resize textarea, to avoid such behavior 
image

2. When we scroll out menu, it overlaps the header

image

AndrMar1939:
> 1. Let's disable ability to resize textarea, to avoid such behavior
> 
> image
> 2. When we scroll our menu, it overlaps the header
> 
> image

dffUqp (approved):
GJ🔥

PR #382, opened 2023-07-26
first version

valet2102:


dffUqp (changes requested):
Looks good:smiley: but always can be better

- let's disable ability to resize the textarea 
- button bigger than your inputs please check design and fix it

image

Anna-Tkachenko (approved):
Good job!
- address link near the contact form should be opened in the new tab
- will be better to limit the image sizes on the medium screen (https://prnt.sc/ay6qhwRAIsKh)

PR #387, opened 2023-07-27
add first task solution

viktoria-roik:
[DEMO LINK](https://viktoria-roik.github.io/layout_landing-page/)


Anna-Tkachenko (changes requested):
Good job! Some comments:
- I can view other elements when menu is opened via `tab` button, let's fix it (https://prnt.sc/F9KaLEVAfgPA)
- all fields in the form should be required. For now, when I try to submit an empty form, I don't see any error. Check form functionality in the checklist
- Will be better to limit image sized on the small - medium screens (https://prnt.sc/DMrhti05jguO)

tiserett (approved):
good job 🔥🔥

I have few suggestions, but I won't reject because of them

- Preferably to remove hover effects on tablet + phone
- Let's decrease scale hover effect just a little bit
- 'About us' is a link to the top of the page, it shouldn't be like this

PR #389, opened 2023-07-27
mobile + tablet layout, desktop in process

polinavafik:
[DEMO LINK](https://polinavafik.github.io/layout_landing-page/)


polinavafik:
I need to fix styles in the footer, so it would look the same as the design, I'm working on it now 👍

polinavafik:
there is a problem idk how to fix
in top-bar there is 
```
&__logo {
    height: 16px;

    @include hover(transform, scale(1.2));

    @include onTablet {
      height: 24px;
    }
  }
```

which makes the logo grow in size on hover. The problem is - it works as expected in aside, but on the main page, it's  bigger than it should be, I don't understand why

Anna-Tkachenko (changes requested):
Let's move all tech questions to q&a chat, mentors won't add comments `how to fix` in github conversation
Some comments about current demo link:
- check logo displaying (https://prnt.sc/CVqY8pjXq9FI)
- disable page scrolling when menu is opened (check point 11 in checklist)
- phone, email and address near the contact form should be the links
- will be better to disable ability to resize for textarea field
- form submitting redirects me to 404 error page, check checklist for correct form functionality

dffUqp (approved):
LGTM🔥

PR #390, opened 2023-07-27
Layout Landing Page My Bike

vserhiichuk:
[DEMO LINK](https://vserhiichuk.github.io/layout_landing-page/)


BudnikOleksii (changes requested):
Looks great🚀
Few more things to fix in accordance to the checklist:
4. Change text color on hover for phone, email and address.
5. Make sure phone icons and phone numbers are real link to start a call.
6. All addresses should be links to some location on Google Maps. Open then in a new tab with target="_blank".
image
image
15. Page shouldn't be reloaded on form submit, but the fields should be cleared. Use:
image

Anna-Tkachenko (approved):
Some details:
- check default font styles for textarea placeholder
- all form fields should be required (check message one)

PR #391, opened 2023-07-27
header and menu added

tillorn:
-[DEMO LINK](https://tillorn.github.io/layout_landing-page/)


Viktor-Kost (changes requested):
Nice!
Keep going!

### To improve: 
1) provide a possibility to call by clicking at this link
Screenshot 2023-07-27 at 20 26 03

**Highly recommended**
As an example, let's take your header background
- Use different images for different screen sizes (mobile, tablet, desktop) using media-queries. You can download these images from figma. We don't want to use big image file size from desktop for mobile. It's ruining your website experience. [Here](https://www.theessential.agency/blog/are-images-ruining-your-website-experience) you can find why is it important
- Use optimized images. [Here](https://squoosh.app/) is an app that can help you to compress your images. Imagine that user have 3g connection and open your website with a phone. 6mb is too big to download

PR #395, opened 2023-07-27
add task solution for Nothing Landing Page by Philip Averenkov

ave-phil:
[DEMO LINK](https://ave-phil.github.io/layout_landing-page/)


Anna-Tkachenko (changes requested):
Good job!
Some comments:
- Add anchor links for the items in the opened menu: each link in the menu should redirect to some section
- It's not clear why the product price redirect to the top of the page for now
- Same with images: it's not clear why it redirects me to the top by clicking on them. Let's just leave hover effect for images but don't redirect user to some other section
- Check form displaying according to the design (https://prnt.sc/dQVolgv0P1pc)
- Google address near the contact form should be opened in the new tab

PR #397, opened 2023-07-27
add task solution

OlhaMomot:
[DEMO LINK](https://olhamomot.github.io/layout_landing-page/#)


Anna-Tkachenko (changes requested):
Good job! Some comments:
- will be better to decrease hover effect a little (https://prnt.sc/NJkFsJwmvo_p)
- all fields in the form should be required. For now when I try to submit an empty form, I don't see any error. Check correct form processing in the checklist

Anna-Tkachenko (approved):
Good job!
Will be better to add some smooth hover effects for the phone links in the opened menu

PR #399, opened 2023-07-27
landing-page

vadimkondratiuk:
[DEMO LINK](https://<vadimkondratiuk>.github.io/layout_landing-page/)


SanyaBratashchuk (changes requested):
1. Add hover effects for images
2. Add favicon
3. Fix fonts 
image
image
4. should be  links 
image
5. all links should be required 
6. form submit shouldn't trigger page refresh

etojeDenys (changes requested):
your link does not work, feel free to ask it in the chat

DarkMistyRoom (changes requested):
- make it a `tel` link
image

- add a hover effect to all images

- in my opinion, would be better to make the scale smaller to avoid text overlapping
image

- make the cursor a pointer over the button
image

- add hover effects to these links
image

- make the address a link to google maps
- don't reload the page on the form submitting
- please fix it
image

- there is no right indent on mobile
image

- add scroll to the menu to make it works fine in landscape mode

etojeDenys (changes requested):
Disable page scrolling under the menu

etojeDenys (approved):
lgtm

PR #405, opened 2023-07-27
add task solution

desevoker:
**NOTHING**

[DEMO LINK](https://desevoker.github.io/layout_landing-page/)


dffUqp (commented):
LGTM😀 but need to fix some points from the [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md)


15. Page shouldn't be reloaded on form submit, but the fields should be cleared.
image

dffUqp (changes requested):
👀

desevoker:
Thanks for the review. The page was unfinished at that moment. Now everything should be okay.

Anna-Tkachenko (approved):
Good job!
Some comments:
- Will be better to limit the size for the images on the medium-small screens, for now looks too large (https://prnt.sc/JZIrMEXnRx2W)

PR #408, opened 2023-07-27
Landing Page MyBiKE

vatatan:
- [DEMO LINK](https://vatatan.github.io/layout_landing-page/)


BudnikOleksii (changes requested):
Great job🔥
Don't forget to add the "Contact Us" section form.
Also, let's improve a bit in accordance with the checklist:
1. Use logo as a favicon
12. Make it possible to scroll the menu if it is higher than the viewport:
image
And all things for forms in accordance with the checklist.

vatatan:
Thanks for the feedback
I am currently working on an add-on

Could you please help me with the solution?
I wrote a mixin in this code for the effect of increasing the picture
But it works for all, except for pictures with bikes, although I also entered it there. The effect doesn't work and I cannot understand why

I will be grateful for your help
thank you

vatatan:
Thanks for the feedback
I am currently working on an add-on

Could you please help me with the solution?
I wrote a mixin in this code for the effect of increasing the picture
But it works for all, except for pictures with bicycles, although I also entered it there. The effect doesn't work and I cannot understand why

I will be grateful for your help
thank you

PR #410, opened 2023-07-27
Not finale version

Wesses:
[DEMO LINK](https://Wesses.github.io/layout_landing-page/)
Still in progress, please, help me to improve my code)


BudnikOleksii (changes requested):
Looks good!
Some things that need to be improved in accordance to the checklist:
1. Use logo as a favicon
4. Change text color on hover for phone, email and address
5. Make sure phone icons and phone numbers are real link to start a call
6. All addresses should be links to some location on Google Maps. Open then in a new tab with target="_blank".
image

src/index.html
```diff
@@ -1,13 +1,253 @@
...
+            </a>
+            <div class="top-bar__icons">
+              <a href="#login" class="icon icon--phone"></a>
+              <a href="#menu" class="icon icon--menu"></a>
+            </div>
+          </div>
+          <div class="header__bottom">
+            <h1 class="header__title">
+              Bring joy back
+              <br>
```
FrontKid, line 37: It would be better to modify the title with css only

FrontKid (changes requested):

PR #413, opened 2023-07-28
add task solution

Orpheus29:
[DEMO LINK](https://Orpheus29.github.io/layout_landing-page/)


Anna-Tkachenko (changes requested):
Good job! Some comments:
- check page title, there should not be `new version` text
- header should have 100vh height, check it
- page scrolling should be disabled while menu is opened (check point 11 in checklist)
- I can also see other elements via `tab` button when menu is opened, fix it
- add some styles on hover state for all images (check point 7 in checklist)
- phone, email and address near the contact form should be the links
- check correct placeholder position (https://prnt.sc/6d7Jtz56krHi)
- all fields in the form should be required, I should not be able to send empty form
- when I click on `submit` button - I should not be redirected to the top of the page, form just should be cleared

Anna-Tkachenko (approved):
Well Done! ⭐️

PR #415, opened 2023-07-28
Develop

bogemna:


Anna-Tkachenko (changes requested):
Add demo link to pull request description

Moroz-Dmytro (changes requested):
As I understand the work is not finished yet
I don't see "footer" and "contact us" sections
But suggestions about what is already done:

1. image

2. Links in menu should scroll the page to some section on the page
3. When we hover on "phone" icon we should see phone number
image

4. Images on bigger screens looks not really good
image
5. 
image
6. This part looks not as on design
image
7. Add some hover effects on the image hover
Go through all points in [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md)

dffUqp (commented):
- you should limit width for this (small tip: try to add align items for you `nav`)

image

- make it possible to scroll the menu if it is higher than the viewport

- remove scroll when menu is open

- make this phone number a real link to start a call

image

- please add some hover effects on image hover

dffUqp (changes requested):
👀

PR #419, opened 2023-07-28
add task solution

voronine:


BudnikOleksii (changes requested):
Please, add the link for the demo so reviewers can test your app.

Anna-Tkachenko (changes requested):
Good job! Some comments:
- add correct favicon to your page (something related to the content of your landing, not react icon)
- page scrolling should be disabled when menu is opened (check point 11 in checklist)
- check correct position for the main title in the header, it should be centered
- add some styles on hover state for all images (check point 7 in checklist)
- address near the contact form should be the link to google maps
- all form fields should be required, for now when I try to submit empty form, I don't see any errors

Anna-Tkachenko (changes requested):
- check favicon for your page
- I can see other elements when menu is opened via `tab` button, let's fix it

qoa11a (changes requested):
I have a 404 error when trying to open the demo link. Please, deploy your solution.

BudnikOleksii (approved):
Nicely done🚀
For better user experience I think hovered elements should be on top of other content, not like this:
image

PR #420, opened 2023-07-28
Develop

edgergard:
[DEMO LINK](https://edgergard.github.io/layout_landing-page/)


Anna-Tkachenko (changes requested):
Did you manage how to fix the demo link? As we don't have the ability to check last version for now
To make it works, just try to run the deploy of the last version one more time

Anna-Tkachenko (approved):
- check font styles for textarea placeholder
- will be better to show the phone near the phone icon in header
- will be better to limit the size for the images on the medium screens (https://prnt.sc/RcPF_bzFwaRi)

PR #429, opened 2023-07-29
Develop

cty2802:
- Deploy and create a Pull Request with a [DEMO LINK](https:///cty2802.github.io/layout_landing-page/)


Anna-Tkachenko (changes requested):
Good job! Some comments:
- add favicon to your page
- for now I see horizontal scroll on the page, it should not exist
- phone number in header should be hidden by default
- all links in the opened menu should redirect user to some section on the page
- let's decrease hover effect for this (https://prnt.sc/SqAyLrsR70Y4) images, for now looks too big
- something is broken with this image (https://prnt.sc/QEfRR4Rf9Ie_)
- placeholder for textarea field is missed
- let's disable ability to resize the textarea field
- check correct form processing according to checklist

PR #430, opened 2023-07-29
add task solution

AngerDESTROYS:


tiserett (changes requested):
- fix adaptive layout
- get rid of hovers on smaller screens (they're redundant)
- phone should always redirect to calling
- phone icon hover - should be able to hover over number (now it disappears
- prevent scrolling on while sidebar is open
- add cursor pointer on every hover
- 'The details' block, rather use same hovers as in 'Compare bikes block
- add favicon + change title

src/index.html
```diff
@@ -1,13 +1,284 @@
+<!-- #region head -->
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" class="page">
   <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Miami</title>
```
alexander-ignatow, line 7: Mind @tiserett comment on title

src/index.html
```diff
@@ -1,13 +1,284 @@
...
+      </section>
+
+      <section class="compare-bikes" id="compare-bikes">
+        <h2 class="section-title">Compare bikes</h2>
+
+        <div class="compare-bikes__bikes">
+          <article class="compare-bikes__product">
+            <img
+            src="images/bike1.png"
+            alt="bike1"
```
alexander-ignatow, line 88: This is a bad value for the `alt`
Please fix other `alt`s too

src/styles/blocks/forms.scss
```diff
@@ -0,0 +1,26 @@
+.input,
+.select,
+.textarea {
```
alexander-ignatow, line 3: It is better to add some dedicated class, like `form-control`, and use it on form elements

src/styles/blocks/forms.scss
```diff
@@ -0,0 +1,26 @@
...
+  border-radius: 40px;
+
+  &:hover {
+    border-color: #d5d7de;
+  }
+
+  &:focus {
+    outline: none;
+    color: #292929;
+    border-color: #292929;
```
alexander-ignatow, line 20: Duplicated colors should be inside variables

src/styles/blocks/section-title.scss
```diff
@@ -0,0 +1,20 @@
+.section-title {
+  margin: 0;
+  margin-top: 90px;
+  margin-bottom: 60px;
```
alexander-ignatow, line 4: No margins for BEM blocks

alexander-ignatow (changes requested):
Not bad! You have a good progress

Still some things to be fixed

1. Please add a favicon as mentioned by @tiserett 
2. Please fix this horizontal scroll - the screen is large enough to fit the content

image

Moroz-Dmytro (changes requested):
- from checklist - 13. All form fields should be required and have correct types (email, tel, etc.)
- Scroll in menu looks not really good(It overlaps header)

image

-  Add some margins from the image on smaller screens(check design)

image

- Menu looks not good in some cases (check screenshot)

image

src/index.html
```diff
@@ -1,13 +1,322 @@
...
+        <div class="compare-bikes__bikes">
+          <article class="compare-bikes__product">
+            <a href="#">
+              <img
+              src="images/bike1.png"
+              alt="first-bike"
+              class="bike__photo"
+              >
+            </a>
+            <h3 class="title">Sporty 4</h3>
```
alexander-ignatow, line 103: Still, alt is not perfect)

I would use the actual bike article - ex. `Bike Sporty 4`

The reason for requiring perfect semantic tags and attributes is that landing is the most critical page from the SEO perspective and should get the most attention from this point

src/styles/blocks/forms.scss
alexander-ignatow: You have missed this one

src/styles/blocks/section-title.scss
alexander-ignatow: This one was missed too

alexander-ignatow (changes requested):
Good 👍 

The only thing is that you have missed a couple of comments

src/index.html
alexander-ignatow: It is not critical for now - so you may leave it as it is :)

BudnikOleksii (approved):
Good job!

PR #437, opened 2023-07-31
add task solution

denlubn:
[DEMO LINK](https://denlubn.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good start 👍 
Let's improve your page
1. Need to change autocomplete styles
image

2. Add `cursor: pointer` for all clickable elements
image

image

4. This image should be the entire width of the window 
image
image

PR #462, opened 2023-08-09
added task layout_product-cards

YevhenPodobedo:
[DEMO LINK](https://YevhenPodobedo.github.io/layout_landing-page/)


YevhenPodobedo:
Good afternoon.
Error correction above described. 
Please pay attention to BEM and what needs to be changed in the semantic tags?
I use correctly preprocessor Sass?

Thank you and waiting form comments.

src/index.html
```diff
@@ -1,13 +1,309 @@
...
+              class="category-block__image"
+              src="images/headphones-hand.png"
+              alt="headphones hand"
+            >
+          </a>
+        </section>
+
+        <h3 class="name category-block__name">All products</h3>
+
+
```
TarasHoliuk, line 179:
```suggestion
```
1 blank line between sibling elements (there should never be more than 1 empty line)

src/index.html
```diff
@@ -1,13 +1,309 @@
...
+          <label for="message"></label>
+          <textarea
+            id="message"
+            class="feedback-form__input feedback-form__input--message"
+            placeholder="Message"
+          ></textarea>
+
+          <label for="btn-feedback"></label>
+          <input
```
TarasHoliuk, line 278: I'd recommend using `button` here

src/index.html
```diff
@@ -1,13 +1,309 @@
...
+
+      
+      <section class="contact-us" id="cart">
+        <h2 class="title contact-us__title">Contact us</h2>
+
+        <form
+        class="feedback-form"
+        method="get"
+        autocomplete="on"
+        >
```
TarasHoliuk, line 253:
```suggestion
        <form
          class="feedback-form"
          method="get"
          autocomplete="on"
        >
```
pay attention to formatting
also check it 15 [here](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md) - might be useful if you will use `button` for Submit

src/styles/page.scss
```diff
@@ -0,0 +1,65 @@
+* {
+  box-sizing: border-box;
+  margin: initial;
+  padding: initial;
```
TarasHoliuk, line 4: do you need it?

src/styles/menu.scss
```diff
@@ -0,0 +1,37 @@
+.menu {
+  box-sizing: border-box;
```
TarasHoliuk, line 2: you already set it in `*` selector - no need to set it anywhere else

src/styles/utils/mixins.scss
```diff
@@ -0,0 +1,205 @@
+@mixin onMobile {
+  @media (max-width: 576px) {
+    @content;
+
+    .header {
```
TarasHoliuk, line 5: you use these mixins a bit unproper

how should it look like:
you have mixin with defined `@media` and with `@content`

then, if you want to add some style to the header, that will be applied only for mobile phones you should do following
```
.header {
  // some styles
  
  @include onMobile {
    // styles you want to be applied for mobile phones
  }
```

TarasHoliuk (changes requested):
From the UI/UX perspective everything looks good
Added comments about code you were asked for :)

src/styles/page.scss
YevhenPodobedo: yes, they effect on the layout (I tried), is it possible to reassign in another place?

src/styles/utils/mixins.scss
YevhenPodobedo: Did I understand correctly?

@include → I use only in the selectors and not @media!  right?
@media → I'm setting a rule, and styles already in the @include.

But! how to be with duplicated @include in the selectors? 

Thanks for the answer.

YevhenPodobedo:
Thanks for the insightful analysis. :-)
Questions arose above....

etojeDenys (changes requested):
few improvements:
1. disable textarea scaling
image



2. should be cursor pointer
image


3. let's import all icons as SVG, because svg icons are scalable, lightweight, customizable, and accessible, providing a consistent and versatile solution for various screen sizes and devices.
image

image

YevhenPodobedo:
will be made! :-)

volodymyr-soltys97 (changes requested):
1. Add transition for hover effects here
image
 
2. Make all these fields are required before sending
image

YevhenPodobedo:
ok! already fixed.

volodymyr-soltys97 (changes requested):
1. Need to change autocomplete styles
image

2. Need to fix font size for phone number from design
image

3. Change title page from `Noting(r)` to `Nothing(R)`
image

YevhenPodobedo:
Oh, my inattention to details is still stuttering....)

I need help! I don't know how to do it yet...
"Need to change autocomplete styles"

tried a lot and not changed styles in the autocomplete! 
In input correction styles.

volodymyr-soltys97 (changes requested):
Still not fixed from the previous review
image

You can read here how to change autocomplete styles - https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/

YevhenPodobedo:
Good afternoon!
I did the right thing?

anastasiiavorobiova (changes requested):
Great work! Almost done! 
Screenshot 2023-08-28 at 13 42 59
Screenshot 2023-08-28 at 13 40 33
This link should be clickable and trigger mail agent etc.

YevhenPodobedo:
done

Anastasiia-Svintsova (changes requested):
- The phone number seems too stretched 
image

- Check the space between the logo and the menu. It doesn't match the design. Check text styles as well 
image

- Doesn't match the design as well 
image

- Check padding 
image
image

PR #476, opened 2023-08-16
solution

ANTI-bogdan:
(https://ANTI-bogdan.github.io/layout_landing-MYBIKE/)


maxim2310 (changes requested):
Nice work, but lets some improvements: 
1. make correct width of this link because now hover effect work on whole line
image
2. hover on phone must show phone number 
image
3. checkout design styles 
image
4. also here and add spaces 
image
5. check out all text styles and all distance between blocks
6. check out your container because now i have horizontal scroll 
image

sTorba24 (changes requested):
Something with your DEMO link, check it please 
image

etojeDenys (changes requested):
1. should be cursor: pointer
image


2. All form fields should be required and have correct types


3. looks different
image
image

etojeDenys (changes requested):
image
feel free to ask in the chat if you have some questions

etojeDenys (changes requested):
image
add required to all inputs

DarkMistyRoom (changes requested):
- check the size of the number. It should be smaller
image

- make the hover effects for these buttons smooth
image

- the container should be 1020px excluded padding. But in your case, it's 1020 included them
image
image

- the button should be wider
image

- fix the font for the texarea
image

- the footer is too big
image

- center the button (tab)
image

- make the menu works fine for a landscape mode. And and some bottom space in the menu
image

DarkMistyRoom (changes requested):
Much better!
- make the hover effect for the button work smoothly
image

- same for these links
image

- fix the font in the textarea. It's different than in the inputs
image

- the footer looks too big. Check its size in the mockup
image

- check the line-height for this section (mobile)
image

- this problem is still here. Ask in the chat if you have questions
image

VitaliyBondarenko1982 (approved):
Looks good to me

PR #486, opened 2023-08-18
implemented header

oleksandrshtonda:
[DEMO LINK](https://oleksandrshtonda.github.io/layout_landing-page/)


TarasHoliuk:
1. See the [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md), point 1
2. Checklist point 8 (and 11, 12):
image
image

Also, implement desktop and tablet versions using @media queries

TarasHoliuk (changes requested):
GJ, Keep going!
Let's add some improvements on the current stage

volodymyr-soltys97 (changes requested):
1. Need to show phone number when hovering this icon
image

2. Add `cursor: pointer` and hover effects for all clickable elements
image

image

oleksandrshtonda:
Answeres:
1: Ok
2: hover affects already exist
3: Ok

anastasiiavorobiova (changes requested):
Looks great! Some comments to fix: 
Screenshot 2023-08-20 at 16 39 42
Screenshot 2023-08-20 at 16 41 40
This link should be clickable
Screenshot 2023-08-20 at 16 41 43
Check font styles here

volodymyr-soltys97 (changes requested):
1. Need to show phone number when hovering only this icon
image

2. Need to fix the background image on the tablet version
Demo
image

Design
image

volodymyr-soltys97 (changes requested):
Still not fixed from the previous review
image

oleksandrshtonda:
I have thought about it for last 3 days and I realized that I can't fix it. I want to explain it.
It requires that height of header should be equal 100vh. In design a device has landscape orientation, but you use screen with portrait orientation. So that's why halfs of wheels hide. 
If we need see all parts of bike when we use portrait landscape, we can remove role "header { height: 100vh}"

mykhalenych (approved):

volodymyr-soltys97 (approved):

PR #503, opened 2023-08-21
Develop

Reject-4444:
https://reject-4444.github.io/layout_landing-page/


Esceype (changes requested):
Great job!!! But here are some issues that still need to be fixed:
Font here:
image
Images should not overlap other images:
image
Wrong indents:
image
Same, there is some block that steal space:
image
There should be an effect of pushing out from under the main content when you reach the footer
image
Here should be placeholder for message:
image

Esceype (changes requested):
Check demo link in your PR description that looks like nothing has changed and the issues is still the same. If on your locale machine everything looks great so try to redeploy project or double-check the DEMO link in description that it looks the same as your locale demo

OleziO (approved):

PR #517, opened 2023-08-26
layout_landing-page

Donovanoff:
[DEMO LINK](https://Donovanoff.github.io/layout_landing-page/)


VitaliyBodnar (changes requested):
Almost done!

- The message field should also be required 
Screenshot 2023-08-26 at 13 40 45

VitaliyBodnar (approved):
Good job!

PR #528, opened 2023-08-30
add task solution

VikiUA:
https://VikiUA.github.io/layout_landing-page/


denys-danyliuk (changes requested):
your demo link doesn't work :(
deploy your changes first

PR #538, opened 2023-09-02
add task landing

IrynaVlasenko:
[DEMO LINK](https://IrynaVlasenko.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Need to remove scroll when this menu is open
image

2. Need to fix the font family for this text and above
image

3. Add transition for hover effects here
image

4. The user should not be able to resize this field
image

5. Need to change autocomplete styles and fix text in textarea
image

nazarmatsevych (approved):
Good work, left two minor comments

1. Add hover effects to this button
image

2.  Add cursor: pointer rule to this button
image

PR #546, opened 2023-09-06
Landing "MyBike"

vashivan:
DEMO LINK](https://vashivan.github.io/layout_landing-page/)


anastasiia-tilikina (changes requested):
Well done! But let's fix some issues though

1. Your pic overlaps another on hover. Consider either increasing scaling a bit or making pic smaller on hover `scale(0.8)` for instance
image

2. You have some default spaces in texarea
image

3. The link `about` in the menu doesn't work

4. Paddings are not applied for the button, check the syntax
image

5. Use the same value for transition duration. It's a good practice.
now your menu opens too slowly.

vashivan:
Thank you for comments, because I could see autofill trouble on my devices and through browsers. I think I coped with it, because I still cannot notice it......
Знімок екрана 2023-09-06 о 21 33 05
Знімок екрана 2023-09-06 о 21 28 03

anastasiia-tilikina (changes requested):
here is your issue with additional spaces in textarea. closing tag must be right after opening one. otherwise HTML considers it as spaces and puts in textarea
image

let's fix it)

also unfortunately I cannot notise autofill effects but you can read [here](https://css-tricks.com/almanac/selectors/a/autofill/) and [here](https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/) about it)

vashivan:
Thank you! I have fixed html and added property to my css for .input. Hope all is ok now:)
Знімок екрана 2023-09-07 о 13 25 06
Знімок екрана 2023-09-07 о 13 25 56

maxim2310 (approved):
Good work!!!

PR #548, opened 2023-09-06
Develop

MarkiMark3:
- [DEMO LINK](https://MarkiMark3.github.io/layout_landing-page/)


Esceype (changes requested):
Nice start🙂

Here are the moments that need to be fixed:
1. Here should be effect on hover with make phone number visible, check design
image

2. Disable ability to resize the textarea
image

3. In all about form. All fields should be required. Submit button should scale smoothly to both cases when mouse enter and when it leave.

4. This button should have hover effects check design
image

5. It looks bad on small devices
image

maxim2310 (changes requested):
Good work!!!
But left a couple of things need to be fixed 
1. need to add hover effect on  this button 
image
2. please remove autocomplete styles 
image

etojeDenys (changes requested):
you forgot to fix that comment
image


feel free to ask for help in chat

mykhalenych (approved):

PR #561, opened 2023-09-11
Landing page task solution

kate0427:
[DEMO LINK](https://kate0427.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. should be the links (and add some hovers)
image


2. add favicon
image



3. number should appear on hover
image



4. the menu looks broken
image



5. Increase all images on hover (make them links to #, if there is no better option)

OleziO (approved):

PR #569, opened 2023-09-12
landing page (in progress)

Drach-Illya:
Please keep in mind that this is just work in progress, further updates will arive in the course of the day.


src/styles/utilities/mixins.scss
```diff
@@ -0,0 +1,52 @@
...
+  @include onDesktop() {
+    max-width: 1020px;
+    padding-inline: 120px;
+    margin-inline: auto;
+  }
+}
+
+.container {
+  @include contentPaddingInline;
+}
```
BetterSol, line 29: Looks like this is not a mixin itself, but more of using mixin. I would suggest moving it to `main.scss` :)

BetterSol (changes requested):
Overall looks good to me :). Good job!

superpooperxxx (changes requested):
Hey, good job🔥

🚨🚨🚨 Add DEMO link in PR description

1. You forgot this
image

2. Logo is pixelated (save in higher resolution in Figma and compress in [TinyPNG](https://tinypng.com)
image

3. Should be centered
image

4. I think footer this way looks slightly weird
image

5. Make this look like in Figma
image

6. Make this `mailto` link
image

7. 🚨Make sure all font weights, font sizes and line height are exactly the same as in Figma
8. This button should look like in Figma and have cursor pointer
image

superpooperxxx (changes requested):
Great job🔥 looks much better. I found a few more things😢

1. Reset outlines and fix bugs in every field (the bug happens because you should use `placeholder` attribute instead of `value`, in textarea delete the text between opening and closing tag)
image

2. When typing the color of inputed text should be darker (check the Figma)
image

superpooperxxx (approved):
🔥🔥🔥

PR #570, opened 2023-09-12
add header and navigation

DenysShurubura:
- Deploy and create a Pull Request with a [DEMO LINK](https://DenysShurubura.github.io/layout_landing-page/)


src/styles/page.scss
```diff
@@ -0,0 +1,26 @@
+.page {
+  font-family: 'Space Grotesk', sans-serif;
+  font-size: 16px;
```
ivaZaiets, line 3: I see you use font-size in html tag with class 'page', so in previous places you can not write font-size: 16px;, because it's size for all tags

ivaZaiets (approved):
Good job! If you want, you can improve your code with my advices! Good luck!

src/styles/icons.scss
```diff
@@ -0,0 +1,17 @@
...
+.icon--phone {
+  background-image: url(../images/icons/phone-icon.png);
+}
+
+.icon--menu {
+  background-image: url(../images/icons/burger-menu-icon.png);
+}
+
+.icon--close {
+  background-image: url(../images/icons/close-icon.png);
```
MolotayMykola, line 16: Do more interactive styles from design like hover effect

MolotayMykola (approved):

polosanya (changes requested):
Good job! Looks great! Here are some moments that would be great to fix:
1. Icons are too small on the desktop. Also, make sure that phone number appears on hover
image
image
2. Prices should be bold
image
image
3. Would be great for this section to take the whole user's screen
image
4. The form should reset on submit, not navigate somewhere.
Also, let's add some validation for inputs (e.g. required, email format)
image
5. Button should take the whole width
image
image
6. Make sure that you add a transition for all hover effects. Also, email should be clickable
image
7. Add some hover effects in burger menu
image
image

src/styles/categories.scss
```diff
@@ -0,0 +1,65 @@
...
+        grid-column: span 4;
+      }
+    }
+  }
+
+  &__photo {
+    width: 100%;
+    height: 100%;
+    object-fit: cover;
+    transition: all 0.3s;
```
ArtemTeslenko, line 51: Checklist p.9 The speed of animations should be the same throughout the page. There is another speed in file button.scss.

src/styles/categories.scss
```diff
@@ -0,0 +1,65 @@
...
+
+    &:hover {
+      transform: scale(1.1);
+    }
+
+  }
+
+  &__title {
+    margin: 0;
```
ArtemTeslenko, line 60: Better to move into reset file.

src/styles/categories.scss
```diff
@@ -0,0 +1,65 @@
...
+    font-weight: 700;
+    font-size: 24px;
+    line-height: 140%;
```
ArtemTeslenko, line 63: You can use shorter variant - 1.4. (optional for you)

superpooperxxx (changes requested):
You are almost there🔥

1. Let's download images in bigger resolution and compress them with TinyPNG (this way they will look better and weigh less)
2. This button should not take the whole container width on desktop
image

3. Let's make navigation appear and disappear smoothly
4. Also links in navigation should lead to a corresponding section

src/styles/nav.scss
```diff
@@ -0,0 +1,38 @@
+.nav {
+  &__list {
+    margin: 0;
+    padding: 0;
+    list-style: none;
```
ArtemTeslenko, line 5: To reset file.

ArtemTeslenko (changes requested):
I leave some comments in your PR. Also 'view all' button in section 'categories' should not occupy the entire width of the parent element on desktop.

superpooperxxx (approved):
🔥🔥🔥

PR #578, opened 2023-09-12
add first part of landing page

Helen-arch:
[DEMO LINK](https://Helen-arch.github.io/layout_landing-page/)


polosanya (changes requested):
Nice! Looks great!
Here are some moments that would be great to improve:
1. Phone number should appear on hover
image
image
2. Prices should be placed in one line
image
image
3. Link titles don't match the design. Also, let's add some hover effects for them
image
image
4. All interactive elements should have cursor pointer on hover
image

vasylmiskiv (approved):

PR #582, opened 2023-09-12
landing first iteration

HardCodWorker:
[demo] (https://hardcodworker.github.io/layout_landing-page/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4727,6 +4727,11 @@
       "integrity": "sha512-+Ryf6g3BKoRc7jfp7ad8tM4TtMiaWvbF/1/sQcZPkkS7ag3D5nMBCe2UfOTONtAkaG0tO0ij3C5Lwmf1EiyjHg==",
       "dev": true
     },
+    "async": {
+      "version": "0.2.10",
+      "resolved": "https://registry.npmjs.org/async/-/async-0.2.10.tgz",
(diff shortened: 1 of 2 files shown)


Drach-Illya (changes requested):
Your header seems to be created in accordance with Misha's instructions. I hope the rest of the site will be coming along just as smoothly once you deploy it.

polosanya (changes requested):
Looks great! Cool animations 🚀 
Here are some moments that would be great to fix:
1. The color of the phone is different
image
image
2. This button does nothing on click. Would be great to navigate somewhere instead
image
3. Make sure that all interactive elements have cursor pointer on hover
image

HardCodWorker:
> 1. The color of the phone is different
To make the phone-number color white was made on purpose, as when it's black it's not vivid.
If it's not crucial I would kindly ask you to let me leave white color

I have fixed other issues according to your review, pleasel let me know if there is anything else that could be improved

polosanya (changes requested):
Nice! Please make sure that you fixed all previous and current moments 
1. Color not fixed
image
2. Typo in "address"
image
3. Images should be square on mobile
image
image

HardCodWorker:
To make the phone-number color white was made on purpose, as when it's black it's not vivid.
If it's not crucial I would kindly ask you to let me leave white color

All other stuff is fixed

superpooperxxx (approved):
Incredible job🔥 I recommend fixing only 1 think

1. Make sure your navigation is vertically scrollable when the content overflows. Here is the example on iPhone SE landscape
image

superpooperxxx (approved):

PR #583, opened 2023-09-12
first landing, contacts and footer will be done soon

YaroslavYarynych:
[DEMO LINK](https://YaroslavYarynych.github.io/layout_landing-page/)


YaroslavYarynych:
I don`t know why, but in demo link my animation doesn`t work and because of it my blocks are hidden. I`ll fix it tomorrow.

src/styles/main.scss
```diff
@@ -1,3 +1,6 @@
-body {
```
ivaZaiets, line 1: Check this place, if you want. I think here you can add some padding, because for me it looks tight :D 
image

src/styles/main.scss
```diff
@@ -1,3 +1,6 @@
...
-  margin: 0;
```
ivaZaiets, line 2: Also in check.list write that color in phone, email etc. should change after hover

image

src/styles/main.scss
```diff
@@ -1,3 +1,6 @@
...
-}
+@import "./utils/variables.scss";
```
ivaZaiets, line 1: And here, also very tight, haha. If you see, image overlaps text. I think you can use scale with less value, maybe 1.05. Try this, if you want

image

src/styles/sections/about.scss
```diff
@@ -0,0 +1,17 @@
...
+    line-height: 32.2px;
+  }
+  @include onTablet {
+    display: grid;
+    grid-template-columns: 1fr 1fr;
+    grid-template-rows: 1fr;
+    justify-content: space-between;
+    gap: 24px;
+    &__text {
+      font-size: 23px;
```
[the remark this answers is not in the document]
YaroslavYarynych: That code is duplicated, thanks! I`ve corrected it

superpooperxxx (changes requested):
Good job🔥 You are almost there
1. I suggest turning off phone number reveal for mobile
2. Page title is located incorrectly according to layout
image

3. Not sure that this link should lead to another page
image

4. 'Move free' is misallocated according to layout
5. Content show slightly late
image

6. This button is too big
image

7. Check the styles in Figma for the form (it should be entered, also borders should be fully rounded, etc)
image

8. Line height here is incorrect
image

superpooperxxx (changes requested):
Great job🔥 Few more things😢

1. I think those icons are too big
image

2. Contacts do not appear on mobile
image

3. Check the styles for the form in Figma
image

Everything else seems to work fine🔥

superpooperxxx (approved):
🔥🔥🔥

PR #592, opened 2023-09-13
add task solution

tilugdmytro:
[DEMO LINK](https://tilugdmytro.github.io/layout_landing-page/)


polosanya (changes requested):
Looks great! Here are some moments that would be great to improve:
1. Background image doesn't match the design
image
image
2. Phone number should appear on hover
image
image
3. Download the image without that button in the corner
image
4. Name and email placeholders don't match the design
image
image
5. Make sure that you added transition for all hover effects. Also, all interactive elements should have cursor pointer on hover
transition
6. Add hover effects for these links
image

tilugdmytro:
Thanks for your comments!)

polosanya (changes requested):
Almost perfect! I'm approving it, but please pay attention to these moments:
1. It would be great to add the ability to stay on the phone number (e.g. to copy it)
phone
2. Transition is still absent for hover effects of buttons
transition

polosanya (approved):

PR #593, opened 2023-09-13
nothing is ready, everything is going to be

Ayogoharo:
[CHECKLIST](./checklist.md) when finished;
[DEMO LINK](https://ayogoharo.github.io/layout_landing-page/)


superpooperxxx (changes requested):
Wow, you almost nailed it🔥
1. You forgot to implement this
image

2. Make cursor pointer for these buttons (it shows that the button is clickable)
image
image

3. Check the font weight and line height in Figma for this
image

4. Lets make these elements link (mailto)
image

5. You should be really careful with line height set in %. Here is why
image

Iphone SE portrait
image

Iphone SE landscape
image

Overall really good job🔥 Please hardcode line height in pixels and maybe think of adding animations, you can use AOS library

superpooperxxx (changes requested):
You nailed it🔥🔥🔥 But I found 2 more bugs

1. Disable horizontal expansion of textarea
image

2. Add inner scroll when navigation doesn't fit to viewport
image

superpooperxxx (approved):
Hey, everything looks good🔥, but form should be sent by clicking on Send button. Please be more attentive to checklist🙏🏻

image

PR #594, opened 2023-09-13
landing-page

tetyana-pol:


DenysShurubura (approved):

superpooperxxx (changes requested):
Good job🔥

1. You forgot this
image

2. You probably forgot to remove this background
image

3. This button should stick to left side of the container
image

4. Check the design for those elements in Figma (they should not be the part of container and lower one should have a background)
image

5. Disable horizontal expansion of textarea
image

6. Let's add cursor pointer to all buttons 
image

7. On landscape mobile layout breaks
image

superpooperxxx (changes requested):
Good job🔥 Looks much better

1. You don't need this margin
image

2. You didn't implement the comment from previous review about adding `cursor: pointer` to all buttons
image

Everything else looks awesome

polosanya (changes requested):
Keep up the wonderful work! Almost perfect 🚀 
1. "Home" is absent on the design
image
image
2. Pay attention to the indentation here. Both titles should have the same position
image
image
3. Would be great to add a validation for email input (set the correct type)
image
4. This part doesn't match the design
image
image
5. Not critical but I think it's better to leave the original title)
image

polosanya (approved):
Well done! I'm approving it, but please pay attention to these small moments)
1. Pay attention that there should be 32px indentation at the top on desktop. Same in burger menu
image
2. Wrong indentation in burger menu here
image
image
3. Let's add transition for all hover effect so all of them work smoothly
hover-transition

4. Check one more time all indentations for mobile, tablet and desktop versions
image
image
5. Make sure that you add a transparent border for the default state of the element and just change its color on hover to avoid such effects:
hover-inputs

PR #605, opened 2023-09-14
Develop

PaninDmytro:
[DEMO LINK](https://PaninDmytro.github.io/layout_landing-page/)


tetyana-pol (approved):

superpooperxxx (changes requested):
Good job🔥

1. I think hero section should take 100vh
image

2. Check if all of your link respect this
image

3. In the mockup we have a gap between rows of text
image

4. Form doesn't look as in design
image

5. Add hover state for this button
image

superpooperxxx (changes requested):
You are almost there🔥

1. Add favicon
image

2. Try quickly hovering between all form field and you will see that those fields jump because of this border (🤫it should always be there, but transparent when you do not hover over)
image

3. Also still font family does not match and outline is not a part of design
image

polosanya (changes requested):
Almost perfect!
1. Icons should be bigger on the desktop. Also, please pay attention to the distance between them
image
image
2. Make sure that you add a transition for the default state of the element, not hover, so numbers appear and disappear smoothly
image
3. Font should be a bit bolder
image
image
4. Would be great to add hover effects for all interactive elements and links
image

StasSokolov (approved):
Awesome job🔥

PR #606, opened 2023-09-14
add task solution

Vitalikmy:
https://vitalikmy.github.io/my-bike/


superpooperxxx (changes requested):
Good job🔥

1. Compress images in order to optimise loading time
2. Styles to not match to Figma
image

3. In design this text is higher (check for all devices)
image

4. In design this text has bigger font size
image

5. In design this text is restricted by the width of the left image
image

6. The text styles in this form doesn't match to design😢

Your solution
image

Design
image

7. This text also doesn't match
image

8. Gap looks too big + you can also make this email a link (mailto)
image

9. Hero text is not centered on mobile
image

🚨If you need any help I am running QnA session today🤗

superpooperxxx:
10. This text should be bold
image

11. Not sure favicon should look stretched 
image

polosanya (changes requested):
Keep up the wonderful work! Almost perfect! 🚀 
1. Better add hover effects for each separate image, not the pair of them
image
2. This button does nothing on click. Would be great to navigate somewhere
image
3. Styles for inputs are different
image
4. Pay attention to the indentations here
image
image
5. This comment is not fixed. Pay attention to the design
image
image

polosanya (approved):
Nice! I'm approving it, but please make sure that you fixed all previous comments
1. Phone number doesn't match the design
image
image
2. This part still doesn't match the design. Pay attention to the font-weight of titles
image
image
3. This button still does nothing on click. Also, make sure that all hover effects work smoothly and have the same transition-duration everywhere
image
4. The indentation here is not fixed
image
image
5. This section should be centred on tablets
image
image

PR #610, opened 2023-09-15
solution

Moblat890:


sTorba24 (changes requested):
Good start! Keep it up
request review when you finish with it

SanyaBratashchuk (changes requested):
image

SanyaBratashchuk (changes requested):
scroll still appears 
image

IvanFesenko (approved):
Well done

PR #629, opened 2023-09-23
solution

killfox93:
https://killfox93.github.io/layout_landing-page/


Andriy-Shymkiv (changes requested):
- in design there is a max-width 1260px, but you set 780:
image
that's why you have some problems like this:
image
image
image
- there should be phone number on hover:
image
image
- you should just reset form values on submit, because now it is 405 error:
image
- add cursor pointer on hovering images and other clickable elements
- add hover effects here:
image

volodymyr-soltys97 (changes requested):
1. Add actual title and favicon on the page
image

2. Need to change the autocomplete styles for inputs
image

3. Need to fix the button width here from design, also remove the default border for button
image

4. You need to import a picture in better quality
Demo
image

Design
image

Viktor-Kost (changes requested):
### Great job!

### To improve: 
when viewport is between 1024 and 1260 there are no padding
Screenshot 2023-09-30 at 09 43 20

volodymyr-soltys97 (changes requested):
You need to make a deploy
image

nazarmatsevych (approved):
Looks awesome, here are small suggestion:

1. Make form fields required
image

PR #639, opened 2023-09-28
first version of a landing

ErnestMk1:
[Demo version](https://ErnestMk1.github.io/layout_landing-page/)


Andriy-Shymkiv (changes requested):
- here should be a phone number on phone icon hover:
image
- nothing happens when I click on phone icon
- add cursor pointer:
image
- decrease transition here and check it everywhere:
image
- add some hover effects for the form elements and details:
image
- nice joke btw but the form shouldn't be updated or redirected on submit:
image

Viktor-Kost (changes requested):
### Nice!

### To improve:
1. remove body scroll when menu is opened
2. make form fields required
3. reset form fields after form submit
4. use a correct image (without this badge)
Screenshot 2023-09-29 at 15 26 55

ErnestMk1:
> ### Nice!
> ### To improve:
> 1. remove body scroll when menu is opened
> 2. make form fields required
> 3. reset form fields after form submit
> 4. use a correct image (without this badge)
> 
> Screenshot 2023-09-29 at 15 26 55



> ### Nice!
> ### To improve:
> 1. remove body scroll when menu is opened
> 2. make form fields required
> 3. reset form fields after form submit
> 4. use a correct image (without this badge)
> 
> Screenshot 2023-09-29 at 15 26 55

"2. make form fields required
> 3. reset form fields after form submit"
Previous reviewer said that on clicking submit the form shouldn't be updated. So, which recommendation should I apply?

ErnestMk1:
The previous reviewer said that on the submit button the form shouldn't be updated. If I add type="submit" for the button, the user will be redirected to nowhere and will have to go back to landing manually. So, which one of your opinions should I follow?

xiimiik (changes requested):
the form should be cleared after submission, and without reloading the page. we don't poison any data, it's just a simulation
image
when the field is empty, a horizontal scroll appears
image
on wide screens on hover images is overlaping each other
image
remove scroll when menu is active
image

etojeDenys (approved):
good job

PR #648, opened 2023-10-01
fix layout

Tvinki228:
https://Tvinki228.github.io/layout_landing-page/


nazarmatsevych (changes requested):
Almost done,  pls check my comments below:

1. Show the phone number on the phone icon hover
image

2. Images should not overlap each other
image

volodymyr-soltys97 (changes requested):
1. Need to fix the background image in footer
Demo
image

Design
image

2. Need to fix the position background image on the tablet version
Demo
image

Design
image

nazarmatsevych (approved):
👍

PR #658, opened 2023-10-06
landing head

Vasyl1991:


denys-danyliuk (changes requested):
Please, provide demo link first

anastasiiavorobiova (changes requested):
Great work! There is no need to change color on the hover here 
Screenshot 2023-11-28 at 16 38 37
So, go ahead! Will be waiting for other sections implemented

PR #662, opened 2023-10-07
The MET

Talpidaeasd:
[DEMO LINK](https://Talpidaeasd.github.io/layout_landing-page/)


VitaliyBondarenko1982 (changes requested):
Good job.
Needs some improvements.
- on hover to phone icon in header should appear pnone numbers
design:
Screenshot from 2023-10-07 15-19-33
actual: 
chrome-capture-2023-9-7 (1)

- add transition for all elements with hover effect (links, buttons, images). transition should be added to element selector (not to selector with hover). it will make smooth  change styles in two ways. 
chrome-capture-2023-9-7
-  [fix autocomplete styles](https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/) for inputs and suggest to make [custom scrollbar](https://www.w3schools.com/howto/howto_css_custom_scrollbar.asp) for textarea
Screenshot from 2023-10-07 15-17-52

etojeDenys (approved):
looks good

PR #663, opened 2023-10-07
add task solution

yursprava:
[DEMO LINK](https://yursprava.github.io/layout_landing-page/)


Andriy-Shymkiv (changes requested):
- something wrong with padding here:
your work:
image
design:
image
- it should be a real phone call:
image
- add hover effects for the links:
image
- add hover for this images and for all others:
image
- change styles for `focus`. you have a design for it:
image
- add hover effects for for items in this section. also it should be a real call on click, real emai and a link for google maps
- remove this phone on mobile:
image
- use svg instead of png:
image

anastasiiavorobiova (changes requested):
Great work! Let's make it perfect! 
Screenshot 2023-12-06 at 13 54 35
Screenshot 2023-12-06 at 13 54 08
Check the font size here 
Screenshot 2023-12-06 at 13 54 01
Could you please add hover styles here? 
Screenshot 2023-12-06 at 13 53 12
This link should be opened in the new tab 
Screenshot 2023-12-06 at 13 53 35
Consider aligning the navigation 
Screenshot 2023-12-06 at 13 52 52
Consider removing the default outline and checking the placeholder for the message field 
Screenshot 2023-12-06 at 13 52 17
A user should be able to scroll the menu if needed 
Screenshot 2023-12-06 at 13 50 53
Check the background-repeat here

yursprava:
1) Consider fixing validation errors

in my code I have href="tel:+123455555555" but not href="tel:+1 234 5555-55-55"

2) Could you please add hover styles here?

there is no examle in the mockup, and my guess is:

&:hover {
      transform: scale(1.05);
    }


3) ...checking the placeholder for the message field

if you press to backspace, you will see a placeholder; I don't know how it works, but every window refresh return this mistake

etojeDenys (changes requested):
1. make all fields required in the form
2. it must be the button instead of the `a` tag 
image

3. set center on desktop looks much better
image
image

DarkMistyRoom (changes requested):
- check the indent between the number and the icon. Looks like it should be bigger
image

- make cursor a pointer over this button
image

- disable the page scroll when the menu is opened and add some bottom indent to the menu to make it look good in landscape mode
image

l4st1m0za (changes requested):
Great job, you are so close!

1) I can't hover on the phone numer, as it's dissapeares as soon as I hover off the phone icon
image

2) Logo should have hover animation 
image

3) I can sent an empty form, please add required fields. 

4) Remove autofill styles for the email field
image

5) Make hover effect in the menue smooth

VitaliyBondarenko1982 (approved):
Looks good. Approve, but want to mention that looks like that still possible to sent form with empty fields inspite of that every input has required attribute.
It is because button has type="reset". Change on submit type. And add such code to form tag ([checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md))
Screenshot from 2023-12-23 15-46-07

You can fix it during preparing project for portfolio.

PR #695, opened 2023-10-17
added landing page

OstinM:


ArtemTeslenko (changes requested):
Good job! There are some issues
1. There is no favicon
2. Wrong positioning of phone number when hover [screen](https://prnt.sc/E4MpCURU5mLX)
3. Page scrolling is enabled when menu is active
4. It is impossible to scroll the menu if it is higher than the viewport
5. There should not be border on inputs [screen](https://prnt.sc/OoV8OPdkFVRC)
6. Try to type in textarea, something went wrong with styles [screen](https://prnt.sc/jaBuprD-1IcS)
7. Do not allow to resize textarea [screen](https://prnt.sc/hYBcxvre4R8h)
8. Address should redirect to google map
9. Click on phone number should start call [screen](https://prnt.sc/hYBcxvre4R8h)

SerhiiKirik:
https://www.loom.com/share/61469a2b44fe4aba8c8a6dfc38301519

ArtemTeslenko (changes requested):
Good job! See Serhii's comments for this request, remove extra styles, add behavior smooth when navigate on page, remove `_blank` from navigation links and phone number, move hover from pseudo class `:hover` to element which should be hovered.

polosanya (changes requested):
Keep up the wonderful work! A few more moments to polish:
1. Not critical but it would be great to add the ability to stay on the phone number for the user. (e.g. if they want to copy the number)
phone
2. Placeholders don't match the design
image
image
3. Let's add validation for email. Also, make sure that there is no error on submit. It would be great just to reset the form on it
image
4. The whole area of the link should be clickable not just the text. It will improve UX. 
links-click

BudnikOleksii (changes requested):
Looks great 🚀 
Just a few moments:
- Checklist. 13. All form fields should be required and have correct types (email, tel, etc.)
- Just email/phone/address should be links, you don't need to do texts links too.
image
- That is just my opinion. You don't need to increase the size of inputs/textarea on hover, it looks strange.
image

BudnikOleksii (changes requested):
I see, why it doesn't work, you set the button type as reset 🤷‍♂️ 
To submit form the button should have type submit.
image

You can reset the form only after submission, otherwise, it doesn't make any sense. You can do it as mentioned in the checklist:
image

BudnikOleksii (approved):
Good job 👍

PR #704, opened 2023-10-19
MYBIKE landing page

VL2808:
- [DEMO LINK] (https://vl2808.github.io/layout_landing-page/)


Andriy-Shymkiv (changes requested):
1. it's better to change cursor for whole page instead of only hover cases
2. please double check the form styles for focus/after cases:
image

etojeDenys (changes requested):
1. you should start the call on click
image

etojeDenys (approved):
awesome!

PR #708, opened 2023-10-20
add landing page

teejrs:
[DEMO LINK](https://teejrs.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
image

2. Need to remove the autocomplete styles

volodymyr-soltys97 (changes requested):
Now you just deleted the autocomplete, this is not the right solution, you just need to stylize this autocomplete, so you need to return it and stylize, you can read how to do it here
image

volodymyr-soltys97 (changes requested):
1. Need to fix the image width here as on the design
image

2. Add a transition when this menu appears on the page
image

3. Need to hide these phone number on the tablet and mobile versions
image

anastasiiavorobiova (changes requested):
You have a great progress here! Will be waiting for implementing a responsive page
Screenshot 2023-10-24 at 15 24 10
Consider fixing the validation error
Screenshot 2023-10-24 at 15 24 24
A user should be able to pick a phone number and call
Screenshot 2023-10-24 at 15 26 24

Screenshot 2023-10-24 at 15 26 04
Check a container width
Screenshot 2023-10-24 at 15 25 47
Check indents here
Screenshot 2023-10-24 at 15 26 40
Consider reducing the element width on hover to avoid overlapping.

PR #710, opened 2023-10-21
add task solution

mykhailoyurkovskyi:
[DEMO LINK] - https://mykhailoyurkovskyi.github.io/layout_landing-page/


polosanya (changes requested):
Nice start! Here are some moments that would be great to fix:
1. Add a favicon
image
2. Scroll should be disabled in burger menu
image
3. Add increasing hover effects to all images. Also, try to download the image without that button in the corner
image
4. All interactive elements should do something on click (e.g. navigate somewhere)
image
5. There should be no error on the form submit. Just reset the form inputs on submit
image
6. Address doesn't match the design
image
image

polosanya (changes requested):
Keep up the wonderful work! 🚀 
A few more comments to fix:
1. The MyBike logo should be the same size in burger menu on desktop
image
2. The burger menu doesn't work correctly on mobile. The horizontal scroll is present and another section is visible
image
3. The mobile version doesn't work correctly at all
image
4. Same with the tablet version
image

BudnikOleksii (approved):
Nicely done 🚀

PR #723, opened 2023-10-24
landing solution

emilionila:
[DEMO LINK](https://emilionila.github.io/layout_landing-page/)

This is not finished version on landing page! While im still working on some features, i would like to recieve a review of existing code ;) .
I ticked below everything i have added:

- [x] Use the logo as a favicon
- [x] Use the landing name as a page title
- [x] All logos should be links to the top of the page
- [ ] Change text color on hover for phone, email, and address
- [x] Make sure phone icons and phone numbers are real links to start a call
- [x] All addresses should be links to some location on Google Maps. Open them in a new tab with target="_blank".
- [ ] Increase all images on hover (make them links to #, if there is no better option)
- [x] Make sure everything looks neat on mobile and without horizontal scrolling
- [ ] The speed of animations should be the same throughout the page (for example, increasing when hovering or moving blocks when scrolling)
- [x] Ensure that all nav__links work smoothly
- [x] Disable page scrolling under the menu using the next code
- [ ] Make it possible to scroll the menu if it is higher than the viewport
- [ ] All form fields should be required and have correct types (email, tel, etc.)
- [x] Placeholders should give examples of what should be added, to make the expected format clear
- [ ] The page shouldn't be reloaded on form submit, but the fields should be cleared. 
- [ ] Make the footer sticky at the bottom with z-index: -1 to get a fixed background effect.


And678 (changes requested):
Looks great! One thing outside the checklist:
Low quality version of icon, a bit blurry, also it doesn't display the phone number.
image

emilionila:
Hi, still working on some tasks, but main things are done. So I would be happy for a feedback ;)
If you have any suggestions on how to stick footer to the bottom , I would be happy to hear them. Also I haven`t increased all the icons versions and I haven`t added phone number that displays on hover yet. Gonna deploy it as soon as will make it done.

emilionila:
Icons now are not blurry

BudnikOleksii (changes requested):
Looks good 🔥 
But, I think this requirement "Make the footer sticky at the bottom with z-index: -1 to get a fixed background effect." is related to other landing pages, because to do something like that with a form doesn't make sense.
Because now I even can't submit a form or click links(phone, email, address). And it says **"to get a fixed background effect"**, but your form doesn't have any background

BudnikOleksii (approved):
🔥

PR #727, opened 2023-10-25
added task solution

Yevhen-Kolosov:
- [DEMO LINK](https://Yevhen-Kolosov.github.io/layout_landing-page/)


roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.
According to the task:
- all phone numbers must be clickable (phone numbers in the menu are not clickable)
- increase all images on hover

BudnikOleksii (approved):
Nicely done 🔥 
Just one moment regarding "View all exhibitions" button, as you have only one page you can use the anchor tag as well, or you can open it in a new window if you want.
And `h5` in this case looks strange, it's just a link
image

PR #728, opened 2023-10-25
1v landing finished

Fluebubble:
- [DEMO LINK](https://fluebubble.github.io/layout_landing-page/)

1. I don't understand that point in checklist.md file, so I didn't relize it:
  - 16. Make the `footer` sticky at the bottom with `z-index: -1` to get a `fixed` background effect.
2. I added image between Categories and About blocks as ::after of "about" block.


polosanya (changes requested):
Nice start! Here are some recommendations:
1. Add the correct title and favicon
image
2. The background color doesn't match the design
image
image
3. All images should have increasing hover effects
image
4. The button text doesn't match the design
image
image
5. Pay attention to the font styles
image
image

Fluebubble:
fixed

BudnikOleksii (approved):
Nicely done 🔥 
Approved, but some of the fonts are still not the same as in design.

Your solution:
image

Design:
image

Fluebubble:
> Nicely done 🔥 Approved, but some of the fonts are still not the same as in design.
> 
> Your solution: image
> 
> Design: image

Something strange...
My figma design:
image

PR #731, opened 2023-10-25
Result MyBike Landing

errisi:
[DEMO LINK](https://errisi.github.io/layout_landing-page/)


polosanya (changes requested):
Nice start! Here are some comments that would be great to fix:
1. The phone number should be clickable. Same with icon
image
2. Try to download the image without that button in the corner
image
3. Make sure that all interactive elements have cursor pointer on hover. The current cursor looks okay but it's better to change the default one completely if you want to leave it
4. It would be great to reset the form on submit. Also, let's add some validation (required, email format)
image
5. Pay attention to the font styles here
image
image
6. There should be no scroll in burger menu
image

BudnikOleksii (changes requested):
Good work 👍 
Some suggestions that should be made:
- all interactive elements like buttons/links/icons with some actions, when you hover them, the cursor should become a pointer
- the "Explore" button should be a link too
- checklist 12. Make it possible to scroll the menu if it is higher than the viewport;

image

image

BudnikOleksii (approved):
Good job 👍

PR #739, opened 2023-10-26
First landing page try

Mantserov:
[DEMO LINK](https://Mantserov.github.io/layout_landing-page/)

Making this pull request just to show that I've done at least something with this landing task.. but it's really hard for me and I know there's a lot of troubles, especially with the code structure. A lot of garbage.. but I really don’t know how to get rid of it so that the landing page is at least somehow similar to what it should be..


BudnikOleksii (changes requested):
Good start 👍 
Some features are not implemented, pay attention to the [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md):
- checklist 1. Use the logo as a favicon
- checklist 4. Change text color on hover for phone, email, and address(now they don't have any animations at all)
- checklist 5. Make sure phone icons and phone numbers are real links to start a call
- when a user hovers over the phone icon you should show a phone number.
image
- checklist 6. All addresses should be links to some location on Google Maps. Open them in a new tab with target="_blank".
- checklist 7. Increase all images on hover (make them links to #, if there is no better option)
- on mobile, some sections looks not the same as in design, for example:
image
image
- checklist 11, 12:
image
- 13 - 16 from the checklist are not implemented either
image

PR #740, opened 2023-10-26
Develop

YehorHeneralov:
https://YehorHeneralov.github.io/layout_landing-page/


BudnikOleksii (changes requested):
Good start 👍 
Some requirements need to be implemented:
- Use the logo as a favicon
- When a user hovers over the phone icon you should show the phone number
image
- When I click the "Send" button nothing happens(inputs, textarea, and button should be inside form element).
image
- 15, 16 from the [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md)
image

BudnikOleksii (changes requested):
Looks good 👍 
Just one moment. The background image in the footer doesn't cover all screen width. 
image

artemforsoff:
I checked the latest fix. Works fine on all screens.
The only thing I noticed and could improve on is removing an unnecessary block.
From:
`<footer class="footer"> <div class="footer__img"></div> </footer>`
To:
`<footer class="footer"></footer>`
 I checked that its absence has no effect on anything.

artemforsoff (approved):
The last remark has been corrected, it works correctly.

PR #741, opened 2023-10-27
Develop

mberezhna:


polosanya (changes requested):
Please don't forget to add the demo link to the PR description, so we can properly check your work
image

mberezhna:
https://mberezhna.github.io/layout_landing-page/

polosanya (changes requested):
Nice start! Here are some moments that would be great to fix:
1. Add the correct logo and title
image
2. The phone should appear on hover. Also, the phone icon and phone number should be clickable with the ability to call
image
3. It is better to increase the images on hover, not decrease
image
4. The form shouldn't pass the data through the URL on submit. 
Just reset the form on submit
image
5. Indentations are incorrect here
image
image
6. This section doesn't match the design on tablet
image
7. Also, the button should take the whole width 
image

BudnikOleksii (changes requested):
Good work 👍 
But a few points from the checklist should be implemented:
Points 4 - 6
image
And points 11 - 12.
image

BudnikOleksii (approved):
Great 👍 
Approved, but found one more moment. Disable resize ability for a textarea:
image

PR #744, opened 2023-10-27
add task solution

ikorostylev:
[DEMO LINK](https://ikorostylev.github.io/layout_landing-page/)


denys-danyliuk (changes requested):
add transition for all interactive elements, and don't add it on hover
Знімок екрана 2023-10-27 о 17 08 38

remove ability to resize textarea
Знімок екрана 2023-10-27 о 17 10 27

add background for the logo so this won't happen when menu is higher than viewport
Знімок екрана 2023-10-27 о 17 11 23

also add padding for this last link
Знімок екрана 2023-10-27 о 17 12 04

13'th point in checklist, textarea also should be required

etojeDenys (approved):
lgtm

PR #754, opened 2023-10-29
Develop

olegnovikov10:
[DEMO LINK](https://olegnovikov10.github.io/layout_landing-page/)


BudnikOleksii (changes requested):
Good start 👍 
But a lot of features need to be implemented, pay attention to the [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md):
- Use the logo as a favicon
- 11, 12 point from the checklist:
image
- The burger menu looks strange, especially, if on mobile user rotates the phone:
image
- when a user hovers on phone icon you should show the phone number:
image
- The form looks not the same as on the design:
image
image
- implement other requirements too:
image

PR #757, opened 2023-10-30
MyBikeLanding

Dmytro-Kretov:
[DEMO LINK](https://Dmytro-Kretov.github.io/layout_landing-page/)

### DIFF
--- a/src/styles/blocks/compare.scss
+++ b/src/styles/blocks/compare.scss
@@ -19,7 +19,7 @@
       grid-column: 1 / -1;
       flex-direction: row;
       align-items: baseline;
-      gap: 24px;
+      justify-content: space-between;
       margin-top: 56px;
       padding-bottom: 144px;
     }
@@ -33,11 +33,6 @@
     line-height: 32px;
     letter-spacing: -1.5px;
     text-transform: capitalize;
-
-    &:hover {
-      color: $greyColor;
-      transform: scale(1.1);
-    }
 
     @include onTablet {
       grid-column: 2 / -2;
--- a/src/styles/blocks/contacts.scss
+++ b/src/styles/blocks/contacts.scss
@@ -50,21 +50,20 @@
     &__link {
       text-decoration: none;
 
-      &:hover {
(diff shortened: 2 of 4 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. Need to check the styles for a phone number when hovering this icon
image

2. Need to remove the scroll when this menu is open
image

image

image

5. This link should be opened in the new tab
image

6. Need to fix position for the background image on the footer
image

image

lerastarynets (changes requested):
Pls watch the video review [via the link](https://www.awesomescreenshot.com/video/22026595?key=2e96cd43043fb4d0f838764f0fad071a) :)

Dmytro-Kretov:
> Pls watch the video review [via the link](https://www.awesomescreenshot.com/video/22026595?key=2e96cd43043fb4d0f838764f0fad071a) :)


intoduced changes to the links and added requiered attribute to the form and now it works fine. 
as for bluelish color: frankly, I don't see it in my browser, but neverthelessI have added webkit-autofill to the inputs and I hope now it works fine. 


still, one issue is left. Frankly, I dont know why, the textarea is affecting other inputs and i dont find the solution to it. maybe you could help me with that

IvanFesenko (changes requested):
- Fix this image export from design. No additional elements should be on it
2023-10-30_16h18_40

- Add hover effect to each image, not bot at a time
https://github.com/mate-academy/layout_landing-page/assets/61904995/2cd97d46-b064-4158-ad96-5250bca06d50

- Text area effect other fileds because you add border for it on hover - when you do it size of text area increase

- Form submit lead to 405 page

sTorba24 (changes requested):
On huge width bikes are not centered
image


remove hover from titles(leave it only for links)
image


also you can remove it from section titles(and for `contact us`  you don't have it, would be nice to be consistent in approach)
image

image

sTorba24 (approved):

PR #784, opened 2023-11-10
landing page first commit

baraban2003:
[DEMO LINK] https://baraban2003.github.io/layout_landing-page/

I can't start SCSS. It doesn't start. So I needed to return to CSS as in the video lesson.


SanyaBratashchuk (changes requested):
So far so good. Please re-request review when you finish whole landing

SanyaBratashchuk (changes requested):
1. fix hover state 
image
2. Add some hover effects for these links 
image
3. Add hover effects for images

anastasiiavorobiova (changes requested):
You have a huge progress here! 
Screenshot 2023-11-19 at 18 54 03
Screenshot 2023-11-19 at 18 54 54
Sections overlap on scroll in the landscape mode 
Screenshot 2023-11-19 at 18 52 23
Consider showing the phone number on the hover

volodymyr-soltys97 (changes requested):
image

2. Make all these fields are required before sending, also, when you click on send, the page should not scroll up, and the fields should be cleared
image

3. Need to remove the autocomplete styles for the inputs
image

4. Add a transition for hover effects for buttons
image

5. Need to fix the background image as on the design
Demo
image

Design
image

baraban2003:
//5. Need to fix the background image as on the design Demo
I can't fix it, because the header height is 100%. It looks nice. If I restrict height it will look bad on the desktop.

PR #792, opened 2023-11-13
Add task solution

Poviakalo:
[DEMO LINK](https://Poviakalo.github.io/layout_landing-page/)


maxim2310 (changes requested):

1. need to make scroll smooth after click on anchor link
2. would be nice to remove scroll when menu is opened 
image
3. would be nice to align this prices and add hover on whole card 
image
4. add transition on all element with animation 
image
5. it must be a link and add hover effect
image
6. remove autocomplete styles
image
7. after submit form page must not be crushed 
8. all inputs must be required and email input validate like email
9.

IvanFesenko (approved):
Well done

PR #798, opened 2023-11-15
add task solution

SofiiaYarymovych:
https://SofiiaYarymovych.github.io/layout_landing-page/


denys-danyliuk (changes requested):
Good start! Send this landing on a review after you'll finish it

you should pick image without this border from the design
Знімок екрана 2023-11-15 о 16 22 00

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. The user will not be able to change the size of this field
image

image

image

4. Need to remove the scroll when this menu is open
image

5. Add transition for hover effects everywhere
image

6. Add a placeholder text for the text area
image

7. Need to remove the autocomplete styles for the inputs
image

denys-danyliuk (changes requested):
logo should be bigger and h1 should have bigger font-size, double check your landing, check if you've used right sizes and indents for all resolutions
Знімок екрана 2023-12-07 о 17 36 51
Знімок екрана 2023-12-07 о 17 37 14

this link lead to some other address, also pay attention to 6'th point in checklist

images should be implemented as links, 7'th point in checklist
Знімок екрана 2023-12-07 о 17 41 23

this nav_links lead to the same section, would be better if one of them would lead to some other section
Знімок екрана 2023-12-07 о 17 42 26

add paddings for the last link when menu is higher than viewport
Знімок екрана 2023-12-07 о 17 43 14

etojeDenys (approved):
lgtm

PR #812, opened 2023-11-18
Develop

yurovych:
https://yurovych.github.io/layout_landing-page/
i got  a remark about weird background of top-bar in menu section
So is was requied by ckecklist
there was a part of code that I had to use in mine fo creating this fixed top-bar with another background


loralevitska (changes requested):
Keep up the good work! Some comments to fix:
- make this number reachable 
image
- fix this according design 
image
- make all form field required
- disable form resize, it breaks design
image
- fix image overlap. add div wrapper to image and set it the `overflow: hidden` property
image

PR #820, opened 2023-11-20
prod

andriy-fesych:
[DEMO LINK](https://andriy-fesych.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. disable the scroll when menu is open
image


2. make all fields in the form required
image

3. looks good, keep going

anastasiiavorobiova (changes requested):
Great start! 
Screenshot 2023-11-21 at 14 55 11
Screenshot 2023-11-21 at 14 56 14
* Check the sizes of your elements
* The menu should be scrollable if needed

andriy-fesych:
> The menu should be scrollable if needed

Fixed. Thank You for mentioning.

> Check the sizes of your elements

Sizes are according to mobile mockup. For up to 744px mobile mockup applied. 
[Here is the link to original mock up](https://www.figma.com/file/NZQAIydtHo5QkINyGLHNcq/BIKE-New-Version?type=design&node-id=0-1&mode=design&t=6vcsrKQje2KE0bNY-0)

andriy-fesych:
image
If You tilt your phone to horizontal orientation, my layout shows exactly how it should look like according to mobile mockup.

etojeDenys (changes requested):
looks good for mobile, keep going

andriy-fesych:
> looks good for mobile, keep going

Can You explain what else You want to be fixed please?

etojeDenys (changes requested):
1. set the height: 100vh to the header
2. we usually don't use vw for the font-size, 'cause it can looks strange on wide screens
image
try to fix that

maxim2310 (changes requested):

1. need to make first block on whole height of page(100vh)
2. if you make relative sizes for something need to set max value, because on wide screen it can be so large
3. also menu must be stretch on whole height 
image
4. need to check font weight here 
image
5. dont know if you want that it looks like, on my opinion better to stretch background and make common border 
image
6. when i hover on text field its jumping a bit 
image
7. need to set min width for prevent this behavior 
image
8. need to remove autofill styles 
image
9. need to add indent here like in design 
image

andriy-fesych:
> Great job overall! Just couple tiny things needs to be fixed before approval)
> 
> 1. need to make first block on whole height of page(100vh)

done: header height set to 100vh.

> 2. if you make relative sizes for something need to set max value, because on wide screen it can be so large

can You please provide an example of such behaviour with resolution mentioned? Checked on ultrawide, couldn`t find an issue. 

> 3. also menu must be stretch on whole height
>    image

Done. Stretched menu (nav) to 100vh. Scroll appears, if content does not fit in 100vh.

> 4. need to check font weight here
>    image

Font weight was off indeed👍. Fixed.

> 5. dont know if you want that it looks like, on my opinion better to stretch background and make common border
>    image

Although, there is no requirement for styling cards, fixed the border as per your recommendation. 

> 6. when i hover on text field its jumping a bit
>    image

Noted 👍. Fixed.

> 7. need to set min width for prevent this behavior
>    image

Noted 👍. Disabled user resizing option.

> 8. need to remove autofill styles
>    image

I can`t reproduce this on my PC. Can You tell me which OS and Browser You used for this?

> 9. need to add indent here like in design
>    image

Added indent.


Feel free to add comments.

IvanFesenko (approved):
Well done

PR #829, opened 2023-11-23
MyBiKE landing by maks-n

ghost:
[DEMO LINK](https://maks-n.github.io/layout_landing-page/)


DarkMistyRoom (changes requested):
Good job! Only a few comments:
- the number should also be a link
image

- fix the textarea font, it looks different than in the other fields
image

- add some bottom space to the menu to make it look good in the landscape mode
image

SanyaBratashchuk (approved):

PR #837, opened 2023-11-26
add task solution

marchukolha:
https://marchukolha.github.io/layout_landing-page/


volodymyr-soltys97 (changes requested):
Good start 👍 
Let's improve your page
1. Add the left and right indents for the content as on the design
image

2. Add a favicon on the page
image

3. Need to show the phone number when hovering this icon
image

4. Add a transition when this menu appears on the page
image

5. Add hover effects for all clickable elements
image

Keep going with other sections

marchukolha:
Thank you. 
1) I made a mobile version. The indents seem to be the same as in the sample mobile version. Please specify which ones.
2) I added the favicon from the Internet because I couldn't find it in the design.
3) There was no such topic in the lesson materials. Maybe it will be later? Because this is only lesson 3 of the module.
4) What exactly does the transition mean?
5) This is the mobile version. The template does not provide a hover effect for the mobile version. Or am I wrong?

SanyaBratashchuk (approved):

marchukolha:
Changed the indents in the header.
Replaced the icons with the required format.
Added hover effect to the menu.

Please check it out)

PR #839, opened 2023-11-29
add task solution

MarkMatiushechko:
- [DEMO LINK](https://MarkMatiushechko.github.io/layout_landing-page/)


src/styles/main.scss
```diff
@@ -1,3 +1,170 @@
...
+  left: 0;
+
+  opacity: 0;
+  pointer-events: none;
+}
+
+.page__menu:target {
+  opacity: 1;
+  pointer-events: all;
+}
```
l4st1m0za, line 28: It's better to use nesting as you have SCSS file which alows it. Try to implement it in all the possible places.

src/styles/main.scss
```diff
@@ -1,3 +1,170 @@
-body {
+// #region page
+.page {
+  font-family: Manrope, sans-serif;
+  color: #1B2129;
```
l4st1m0za, line 4: It's better not to use single colour name, but to use constant which represents the colour.

l4st1m0za (changes requested):
Great job! See the required changes. Also, it's better to use **.svg** instead of **.png** images. Continue working!

MarkMatiushechko:
[DEMO LINK](https://MarkMatiushechko.github.io/layout_landing-page/)

PR #844, opened 2023-11-29
My landing page

Svimpik:
(https://Svimpik.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
image

2. Need to show phone number when hovering this icon
image

3. Need to remove scroll when this menu is open
image

4. The indents on the right and left of all sections must be the same
image

5. Need to fix this block as on the design
Demo
image

Design
image

image

sTorba24 (approved):
Well done! I approve it

but add title page 'Nothing`

PR #853, opened 2023-12-03
add task solution

PavloStelmaschuk:
- [DEMO LINK](https://pavlostelmaschuk.github.io/layout_landing-page/)


anastasiia-tilikina (changes requested):
Continue your work!
As for home section I've noticed a couple of issues:

1. you should limit container width

your demo:
image

figma: 
image

2.  Your title seems to have wrong font-size on desktop.
3. Pls check img position on tablet. It differs from the design
4. menu should appear smothly
5. scroll should be disabled while menu is open.
6. don't forget add hover effects for all interactive elements

VitaliyBondarenko1982 (approved):
Looks good

PR #855, opened 2023-12-04
add task solution

IvanVaverchak:
[DEMO LINK](https://IvanVaverchak.github.io/layout_landing-page/)
Please approve my assignment because the old results are not valid after updating the topic. Here is a link to the approved assignment:  https://github.com/mate-academy/layout_miami/pull/2045


VitaliyBondarenko1982 (approved):
GJ!

PR #873, opened 2023-12-06
add task solution landing demo

Ihor-Prodan:
[DEMO LINK](https://Ihor-Prodan.github.io/layout_landing-page/)


Esceype (changes requested):
GJ!
But there are some things to fix here, but it's also important to keep working on the following sections.
Need to fix in current work:
- add favicon;
- logo should be svg format;
- when menu is open scroll should be disabled;
- phone icon should have same hover effect like in design;
- click on phone should show the default call modal window as it is in menu;
- on landscape mode menu should have scroll;

sTorba24 (changes requested):
Nicely done! Some fixes to do

1. add cursor pointer to all interactive elements
image

2. create containers for content and add max-width, not best idea to shrink all content on huge screens, 
image
image

3. it should trigger call on click
image


4. clear form after submit
image


5. it should be links with hover effect and trigger related actions(mailto, tel ..)
image

lerastarynets (changes requested):
Almost there, just one problem left)
Clicking number in the header shouldn't open new window, it should act just like the phone number in footer
Screenshot at Dec 16 10-10-23

SanyaBratashchuk (approved):

PR #881, opened 2023-12-07
Develop

dyarechuk:


SerhiiKirik (changes requested):
Well done 🔥 
Keep it up 🚀

olya-shyian (changes requested):
A stellar display of talent—impressive!🔥

1. Please add a number on hover
image

2. Remove hover effect on mobile and iPad version, there is no need

image

PR #894, opened 2023-12-08
add task solution

avakiel:
[DEMO LINK](https://avakiel.github.io/layout_landing-page/)


olya-shyian (changes requested):
Nicely done!☘️

1. On big screen this section should be in the center as the next section.
image

also this section should be on the same line, on mockup we have:
image


2. When I click on these phone numbers it should be the ability to call

(Make sure phone icons and phone numbers are real links to start a call from CHECKLIST)
image


3. Please remove this scroll from the right side
image


image

5. All form fields should be required and have correct types (email, tel, etc.)

6. Any hover affects should not work on Ipad or mobile. You can have problem like this
image

7. This is what we have on mobile, the picture should cover the entire viewport.
The same problem we have on ipad too
image

8. It's not ok to have something like this
image

YegorKorenkov (changes requested):
1. This hover looks strange. Pictures overlap each other
image

2. Please check links color
image

3. Please add correct tel href for this icon link
image

src/styles/days.css
```diff
@@ -0,0 +1,147 @@
...
+  display: flex;
+  flex-direction: column;
+  font-weight: 700;
+}
+
+.cloisters__holiday {
+  font-weight: 400;
+}
+
+@media only screen and (min-width: 744px) and (max-width: 1259px) {
```
DanilWeda, line 91: Try writing mixin for all media ( phone, tablet, desktop)

DanilWeda (changes requested):
All good! 
Check small improvements! 
Write mixins for media. 

Check header height, in my case it not 100vh.
image

Add smooth for this button transition 
image

Also, align the header in the menu and standard view.
image
image

Many thanks! Great work!

YegorKorenkov (approved):
Good job! Approve! But please correct these small remarks to finalize your work

1. Button hover missed for mobile and tablet resolutions
image

2. Too big space between this blocks
image

3. When you click on the icon it should fire href with **tel:** link(for tablets and mobile)
image

PR #899, opened 2023-12-10
Develop

kapesha:
https://kapesha.github.io/layout_landing-page/


maxim2310 (changes requested):

1. i must be able to click on these numbers
image
2. would be nice to add animation on opening menu 
3. need to add hover on this links and remove scroll in menu 
image
4. img must not overlap each other 
image
5. add some hover on buttons
image
6. need to add hover on these links 
image
7. e bit hard to understand that this is form, plz add background according design 
image
8. need to remove autofill styles 
image

volodymyr-soltys97 (changes requested):
To improve:
1. Add the left padding for the text
image

2. Now you just turned off the autocomplete, this is the wrong decision, you don't need to turn it off, but you need to style it additionally so that the styles of the autocomplete are the same as when entered manually
image

image

4. Need to remove the scroll when this menu is open
image

etojeDenys (approved):
good job

PR #901, opened 2023-12-11
Done! (code looks awful, but it works)

EdenParadise:
I know that this code is too dirty, there are a lot of space for improvements :/


YegorKorenkov (changes requested):
1. Tel number should be link
image

2.When I click on this icon, it redirects me to a 404 page
image

3. Should be fullheight and without vertical scroll
image

4. Should be links with corresponding hrefs(tel, mail, etc)
image

5. The page shouldn't be reloaded on form submit, but the fields should be cleared.

6. All form fields should be required and have correct types

7. Please check input styles according to figma design
image

PR #947, opened 2024-01-02
add task solution

ilyaivasyk:
[DEMO LINK](https://ilyaivasyk.github.io/layout_landing-page/)


l4st1m0za (changes requested):
Good job, but some changes needed:

1) When I hover on the menue icon, I am also triggering the phone icon animation of the number. These two animations should not be connected to each other
image

2) Add right title and also add favicon
image

3) This button in the menue can have some kind of a animation 
image

4) This block seems to be to small for the desktop mode 
image

5) These bikes should have icreasing animation on the hover
image

6) These photos also should have same hover as previous one 
image

7) This block have some issues with the figma design. See how the border radius is in figma for the 'Message' field and also correct the font size in the button
image

8) I should not be able to send the empty form

9) Remove autocomplite styles on the email field

10) Clear the form when it's sent

11) This block should have animation for the links
image

12) This link should lead to the separate web page with dot in the googple maps
image

TarasHoliuk:
1. Disable scroll when the menu is open (there are hints in [the checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md)):
image

2. This link should be open in the new page (it's general rule for all external links):
image

TarasHoliuk (changes requested):
Well done :+1: 
2 more comments :)

loralevitska (approved):
Good job!

PR #961, opened 2024-01-06
add task solution

MakhmudKerimov:
[DEMO LINK](https://MakhmudKerimov
.github.io/layout_landing-page/)


l4st1m0za (approved):
Awesome job here! The only thing to fix is to remove autofill styles 

https://developer.mozilla.org/en-US/docs/Web/CSS/:autofill

PR #962, opened 2024-01-06
add solution

hardaira:
[DEMO LINK](https://hardaira.github.io/layout_landing-page/)


SanyaBratashchuk (changes requested):
Do not restrict width of your page
 image

Move on to implement tablet and desktop dimensions
 Add favicon

vadiimvooo (changes requested):
In general good job, just few things to fix:
1. Not all hover effects still covered by transition
image
2. This part also have not transition. Besides you should make the liks clickable for phone number and email as well
image

vadiimvooo (changes requested):
You are almost there.
1. You don't need to change height of this block on hover, use appropriate transform property
image
The same situation is for footer block.

TarasHoliuk:
1. I see it looks good on 320px screen (as on the mockup):

image

But it looks not good on a bit different screens.
image

No need to hardcode sizes. You could set picture size in relative units and use flexbox properties to make it look good on any mobile screen.

Since you match design it's optional to improve. But I'd recommend you to do it before adding to the portfolio on the employment stage :slightly_smiling_face: 

2. Disable textarea resize:
Screenshot from 2024-02-20 20-30-44

3. Make the menu appear/disappear smoothly

TarasHoliuk (approved):
Looks good, all previous issues were fixed - Approved :heavy_check_mark: 

Added a few suggestions - you can improve your landing page before adding it to the portfolio (later) :slightly_smiling_face:

PR #965, opened 2024-01-08
add task solution

creonluv:
[DEMO LINK](https://creonluv.github.io/layout_landing-page/)


etojeDenys (approved):
1. number should appear on hover
image

2. better to remove this default value for the textarea
image

3. better to import icons as SVG, it will improve the quality at least
image

PR #987, opened 2024-01-15
add full task solution

reznik-denis:
[DEMO LINK](https://reznik-denis.github.io/layout_landing-page/)

[MYBIKE landing](https://www.figma.com/file/NZQAIydtHo5QkINyGLHNcq/BIKE-New-Version?node-id=0%3A1)


l4st1m0za (changes requested):
Great job here, some things to change are listed below:

1) When I hover on the phone, I can't hover on the phone number which appears
image

2) Your special pointer shows up only on the component which have hover effect. It's just seems strange that on the page pointer is usual but on the elemest it changes to a fancy one

3) When I hover over these images they overlap one next to it, and also don't go through it's container in the left side. When image is being scaled, it should just get bigger and thats all
image

4) It's better just to clear the form, and don't send the information to email, as you get such a thing on your form
image

5) This link should lead to an google maps with this address in it
image

reznik-denis:
1. done
2. any other special pointer doesn't shows in morkup, I make this special pointers for all page, but when we have on hover we don't see ane changes with pointer, and it's doesn't OK I think.
3. done
4. I don't understand about form, when we sumited form all data add to email aplication on user computer with our email address? that's all, I only JS method for cleaning form fields, but he is very complitely and big, I'm not sure that this method need on this site.
5. done

reznik-denis:
2. I add cursor with wite color for all page and black cursor for hover. I think it's better desigion in this situation

reznik-denis:
1. done
2. done

Esceype (changes requested):
Still not fixed:
- scroll should be disabled when menu is open:
image

- disable resize for textarea:
image

If you have any questions how to do it please ask in fe_chat

reznik-denis:
I just redeploy project, now all working and all fixed.

etojeDenys (approved):
awesome

PR #994, opened 2024-01-16
add task solution

yuriiyepikhov:
[DEMO LINK](https://yuriiyepikhov.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. must be links with some hover
image

2. make all fields in the form required
image

sTorba24 (changes requested):
1. which format?
image

2.these links should trigger mailing and address should open location on maps
image

anastasiia-tilikina (approved):
Well done!

I would suggest you to add some hover effects for images and make all hover effects change smooth (add transition)

PR #1001, opened 2024-01-19
Layout landing page "THE MET"

under-a:
[DEMO LINK](https://under-a.github.io/layout_landing-page/)

[x] Use the logo as a favicon
[x] Change text color on hover for phone, email, and address
[x] All addresses should be links to some location on Google Maps. Open them in a new tab with `target="_blank"`.
[x] Increase all images on hover (make them links to #, if there is no better option)
[x] Ensure that all `nav__links` work smoothly
[x] Disable page scrolling under the menu
[x] All form fields should be required and have correct types (`email`, `tel`, etc.)
[x] Placeholders should give examples of what should be added, to make the expected format clear
[x] The page shouldn't be reloaded on form submit, but the fields should be cleared.
[x] Make the `footer` sticky at the bottom with `z-index: -1` to get a `fixed` background effect.


l4st1m0za (changes requested):
Great job, but some changes needed:

1) Block with numbers should be closer and should be reachable. What I mean is that this block do show the phone numbers, but I can't click on them to call, as they dissappear as I hover off the phone icon. 
image

2) You need to add speed to the animation and keep it same at all the page, now you have all animations withous speed, they just occure. 

3) Remove changing the size of the message block. It always should have same size. 
image

4) Remove the scroll in the menue, as it's stated in the checklist

5) Menu layout is broken in the tablet and phone mode
image

VitaliyBondarenko1982 (changes requested):
Looks good. Almost done.
Suggest a few improvements.
- I can't to click on phone number link and start a call. For fix it do not make phone icon as a link, place block with phones links inside phone icon element and position absolutely relative to phone icon.
chrome-capture-2024-1-20 (1)
- implement 11 point from [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md)
Screenshot from 2024-01-20 15-49-28
- add cursor pointer for send button. check other interactive elements.

under-a:
> Looks good. Almost done. Suggest a few improvements.
My bad, just forgot to deploy

VitaliyBondarenko1982 (approved):
Well done

PR #1010, opened 2024-01-21
Develop my first landing

boikoua:
- [MYBIKE landing](https://www.figma.com/file/NZQAIydtHo5QkINyGLHNcq/BIKE-New-Version?node-id=0%3A1);
- **DON'T** try to do it `Pixel Perfect` - implement it the most `simple` way so it looks similar;
- When done check yourself using the [CHECKLIST](./checklist.md) when finished;
- Deploy and create a Pull Request with a [DEMO LINK](https://boikoua.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
image

2. Add the transition when this menu appears on the page
image

3. Need to change the autocomplete style for inputs
image

image

5. The user will not be able to change the size of this field
image

6. Add transition for these hover effects
image

l4st1m0za (changes requested):
Great job! Only some fixes needed:

1) Here you have 3 same bikes on photos
image


2) You still have not changed the atuocomplite styles from the previous review 
image

maxim2310 (changes requested):
almost done, just check this point
1. i must be able to click on number and also check font on number
image
2. img must not overlap near one 
image

boikoua:
> 1. i must be able to click on number and also check font on number

Безымянный

Unfortunately there is no way to make it a link so that it would appear and not disappear again, even with the help of mentors. And why should the user click on the number that appears, if he can just click on the phone icon)?

l4st1m0za (approved):
Great job!

PR #1023, opened 2024-01-24
add html & scss with header and menu-bar

helenashalamova:
- [DEMO LINK](https://helenashalamova.github.io/layout_landing-page/)


l4st1m0za (changes requested):
Good start! Call menthor for a next review when your whole page would be done!

PR #1025, opened 2024-01-24
Develop

myehor:
[DEMO LINK](https://myehor.github.io/layout_landing-page/)


DarkMistyRoom (changes requested):
Good job! A few small changes are needed:
- the phone icon should be a tel link and phone numbers should be reachable, because now when I try to reach them they disappear
image

- would be great to add some hover effects to images
- the contacts should be links
image

- the form should fit 5 columns
image
image

- disable the textarea resizing
image

- make the message field also required
- 
image
add scroll to the menu to make it work fine in landscape mode

myehor:
Help me please with header. A can`t decide which is the best way to add block with numbers to phone-icon.

sTorba24 (changes requested):
Great work!

Some hints how to implement show mobile

create block that will work as expected, then hide it and show on icon hover

if you have any questions better to text your fe_chat, because here we can see your comment only after review request and it could by other mentor, so not productive :(

also add some hover effect(color change maybe) to the links
https://github.com/mate-academy/layout_landing-page/pull/1025#pullrequestreview-1854270447

and don't forget that address should be link to google maps place

maxim2310 (changes requested):

1. its also must be a links 
image
2. need to remove autofill styles 
image
3. would be nice to add hover on img
image
4. need to remove scroll when menu is opened

anastasiiavorobiova (approved):
Great work!

PR #1036, opened 2024-01-29
add task solution

styopochkin23:
https://styopochkin23.github.io/layout_landing-page/


volodymyr-soltys97 (changes requested):
Hi, your demo link doesn't work, you need to make a deploy
image

vadiimvooo (changes requested):
Good job, but there is a few improvements. You didn't meet some requirements from previous review:
1. Make sure phone icons and phone numbers are real links to start a call. 
Somewhere it works but not in all places (for example if you try to click in header it will not work and redirect you on 404 page.
2. Remove please redirection to google maps from email in footer and instead of it use mailto.
3. Here is some problems with gaps. Please check the design and make changes.
image
4. This block is not the same as in design.
image
5. Tablet and mobile desing are broken at all, please fix it.
image
6. Add navigation for links which located in burger menu.

PR #1040, opened 2024-01-30
Develop

illya-onyshchuk:
[DEMO LINK](https://illya-onyshchuk.github.io/layout_landing-page/)


BudnikOleksii (changes requested):
Good start, but the most of sections and features aren't implemented

BudnikOleksii (changes requested):
Good progress, some fixes that should be done:
- Use the logo as a favicon
- Menu has a grey background on top and links are uppercased instead of capitalized, check the design:
image
- "View all" button is too big on the desktop:
image
- The form has different design in figma
- When a user hovers phone icon, the phone number should be shown
image

PR #1045, opened 2024-01-30
https://Pavlo-lab-max-Kopalko.github.io/layout_landing-page/

Pavlo-lab-max-Kopalko:


DarkMistyRoom (changes requested):
Good job! But here is some comments:
- please provide the link in the PR description
image

- show the number on hover
image

- avoid overlapping on hover
image

- make the cursor a pointer over buttons. And remove the default outline
image

- would be great to add some hover effect to these links
image

- remove size-changing from the textarea
image

- remove horizontal scroll from tab and mobile screens

vadiimvooo (changes requested):
Good job. A few changes needs to be done:
1. Something is wrong with phone number on hovering the phone icon on desctop and tablet design (I will recommend you to use pseudo element before for this purpose):
image
2. The overlapping problem with hover still exists.
image
3. It will be also nice to have transition effect on color change for your buttons around the app.
4. Remove scroll when burger menu is opened.
image

Pavlo-lab-max-Kopalko:
I left new reference above.

IvanFesenko (approved):
Well done

PR #1047, opened 2024-01-31
layout-landing-page

svyatoslavPy:
[DEMO LINK](https://svyatoslavPy.github.io/layout_landing-page/)


sTorba24 (changes requested):
Your demo link isn't working :( Maybe you forgot to deploy ?

vadiimvooo (changes requested):
Well done! There are a few changes you need to do:
1. Add transition effect on hovering the phone icon to make phone number appearance smothly
image
2. Make this phone number clickable as well as the link below it. 
image
3. This block should be fixed. Please look at the mockup (desktop size).
image
4. It would be good to make this images scalable on hover.
image
5. The same is for this images:
image
6. Here are some problems with header you also need to fix (mobile version):
image
7. Height of this block is not the same as on mockup. Please make changes to it:
image

PR #1052, opened 2024-02-01
add solution

vlesia:
-[DEMO LINK](https://vlesia.github.io/layout_landing-page/)


polosanya (changes requested):
Nice start! Keep working 🚀 
1. Let's add a favicon and the correct title
image
2. Main page doesn't match the design. Check all the sizes
image
3. Phone number should appear here on hover. Also, would be great to add hover effects to all interactive elements
image
4. There should be no scroll in burger menu
image

Moroz-Dmytro (changes requested):
1. Try to find better favicon, this one is not really good

image

2. Hover effect on phone icon is really strange(I should see phone number after hover)


https://github.com/mate-academy/layout_landing-page/assets/86789161/2a1693ff-f467-4626-9a3d-1fa223d05414

3. Images should not overlap each other on hover

image

4. FIx this

image

5. All inputs should be mandatory

image

6. Map should be opened in the new tab after link click

image

7. Add paddings in header on small screens

image

roman-mirzoian (changes requested):
Nice fixes, but still some design differences:
- on the design, this picture looks in a different proportion and the text on it differs in font and size (font issue in all screen sizes):
image
- this indent on the design is less:
image
- Change text color on hover for phone, email, and address (as specified in the checklist)
- All form fields should have correct types (currently, you can enter any value in the email)

BudnikOleksii (approved):
Good work 👍

PR #1053, opened 2024-02-01
Develop

mikeross0104:
[DEMO LINK](https://mikeross0104.github.io/layout_landing-page/)


IvanFesenko (changes requested):
Good job. Check my comments:
- Add favIcon
- Center content
2024-02-01_14h20_55
2024-02-01_14h21_13
- Remove autocomplete styles
2024-02-01_14h21_54
- Form should clear on submit without page reload
- Add hover effect to images and links in footer
- Disable scroll in case menu is open

volodymyr-soltys97 (changes requested):
To improve:
1. Need to remove the scroll when this menu is open
image

2. Need to size for these images as on the design
image

3. Now you just turned off autocomplete, but this is the wrong decision, you don’t need to turn it off, but you need to stylize it additionally, you can read how to do it here
image

4. After clicking send, the page should not reload and scroll up, you only need to clear the fields and that's it
image

TarasHoliuk (approved):
Hi! GJ - Approved :+1: 

Added one comment. You may add changes before adding the landing page to the portfolio, before the employment stage to make it even better :slightly_smiling_face:

PR #1055, opened 2024-02-02
check footer with mentor

troleksii:
[DEMO LINK](https://troleksii.github.io/layout_landing-page/)


l4st1m0za (changes requested):
Your whole web page is broken, may be you have forgotten to make a deploy, but now, it seems to be wrong at all
image
image
image

PR #1056, opened 2024-02-02
Develop

Lida-Petrushynets:
[DEMO LINK](https://Lida-Petrushynets.github.io/layout_landing-page/)


Andriy-Shymkiv (changes requested):
good start!

to improve:
1. use svg images instead of png
2. it should be a real phone call on phone-icon click

PR #1057, opened 2024-02-02
completed version

ViktorTsupa:
https://ViktorTsupa.github.io/layout_landing-page/


maxim2310 (changes requested):

1. optional, now on high resolution we can see this(it will be vertical scroll without dev tools)
image
2. here must be a number on hover that must be reachable 
image
3. scroll must be removed when menu is opened 
image
4. need to remove autofill styles 
image

ViktorTsupa:
did not quite understand point #2, how it should look

l4st1m0za (changes requested):
Good job, but some points should be improved:

1) You steel have autofill styles in email field. You need to make so that your autofill still works, but have the same style as usual text input on your website
https://css-tricks.com/almanac/selectors/a/autofill/

2) This block should also have smooth animation
image

ViktorTsupa:
The first point is unclear to me, prescribing similar options from this site, there are no changes at all, is there a similar example?

SanyaBratashchuk (approved):

PR #1059, opened 2024-02-02
Develop

MOODDDII:
[DEMO LINK](https://MOODDDII.github.io/layout_landing-page/)


vadiimvooo (changes requested):
Hello, here are a few fixes required:
1. The header image should cover all free space:
image
2. You should remove scroll when burger menu is opened:
image
3. The navigation by links in burger menu should be smooth (You can use scroll-behaviour param to implement it.)
4. It would be better if you used hover effect with phone icon and show some phone number on hover:
image
5. On sending form you should prevent page reload.
6. The address line should be clickable and should lead to google maps:
image
7. Please check right round parametr for this field in mockup:
image
8. In general check again all the sizes of elements in the mockup and fix all the sections according to them.

PR #1062, opened 2024-02-03
B&O

OMedvid:
[DEMO LINK](https://OMedvid.github.io/layout_landing-page/)


vadiimvooo (changes requested):
Well done. Here are a few changes to fix: 
1. Add transitions for all elements on your page (buttons and links)
image
image
image
image
2. This font should be bigger:
image
3. Recommended block should be bigger on tablet version: 
image
4. Images should be bigger on phone version:
image

Andriy-Shymkiv (changes requested):
please decrease the transitions. this button is to big to be honest. `1.2` is enough
image
also here:
image

check every element

SanyaBratashchuk (changes requested):
1. apply `cursor: pointer ` for all interactable elements
2. add favicon

TarasHoliuk:
The one thing would be nice to improve before adding to portfolio:
- Show phone number on phone icon hover, as in the design:
image
image

TarasHoliuk (approved):
Approved :heavy_check_mark: 

Added 1 suggestion - you may do it when add project to the portfolio in the future :)

PR #1068, opened 2024-02-05
Develop

anastasia1383:
[DEMO LINK](https://anastasia1383.github.io/layout_landing-page/)


BudnikOleksii (changes requested):
Looks good 👍 
A few moments that should be fixed:
image
image
- On mobile and on tablet section "Recomended" has some problems with centering and paddings:
image
image
- On desktop "View all" button is too big, check the design:
image

Moroz-Dmytro (changes requested):
1. Need to fix

image

2. Recomended blocks should be placed in 1 column

image

BudnikOleksii (approved):
Good work 👍 
Approved, don't forget to do the form fields required, from the checklist:
image

anastasia1383:
Added! Thank you.

PR #1069, opened 2024-02-05
Develop

MOODDDII:
[Demo Link](https://moodddii.github.io/layout_landing-page/)


SanyaBratashchuk (changes requested):
1. add favicon
2. external links should be open in a new tab
3. add `cursor: pointer` for all interactable elements
4. User should be able to submit form. Form should be cleared after submission

PR #1087, opened 2024-02-08
add solution

Graf6612:
https://graf6612.github.io/layout_landing-page/


Moroz-Dmytro (changes requested):
image

3. Number should be shown when you hover on "phone icon"

image

4. "Call to order" color is not as per design

image

5. Hide scroll in menu

image

image

8. Not as per design

image

9. Fix this

image

BudnikOleksii (approved):
Good progress 👍 
Pay attention to the [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md):
- the phone number should be aligned at the center of the phone icon.
image
image
- Make sure phone icons and phone numbers are real links to start a call. Email as well should be a link with mailto.
- All addresses should be links to some location on Google Maps. Open them in a new tab with target="_blank"
- Placeholders in form are different from the design
image
- paddings on mobile are different, they should be the same from the left and right side
image

PR #1096, opened 2024-02-09
Develop

IRONH4MMER:


Moroz-Dmytro (changes requested):
Please, add Demo link to PR description

1. Not possible to hover over phone number


https://github.com/mate-academy/layout_landing-page/assets/86789161/20b49566-441a-461a-a30c-f4aca4ca13b8

2. Main image should be 100vh height

image

3. Images should not overlap each other on hover

image

4. Menu should be also 100vh

image

5. "Book a testdrive should be clickable"

image

6. Images look too small

image

7. Not all menu is visible when rotate phone on smaller screens(and I can't scroll)

image

IRONH4MMER:
1. What for? Phone-icon is already clickable, ::after with phone-number here is only for show;

6. Because it's supposed to be bigger on >=744px (tablet layout).

roman-mirzoian (changes requested):
Good fixed, but it still has some issue:
- there is no link to the demo in the PR description
- according to the task, all pictures and icons must be links 
image

roman-mirzoian (approved):
I see from the changes in the code that the images are now links, but it does not work in the application itself, do not forget to redeploy your application.

PR #1099, opened 2024-02-09
in progress

Vlobdam:
https://vlobdam.github.io/layout_landing-page


Moroz-Dmytro (changes requested):
Need to reject, as it is still in progress

polosanya (changes requested):
Nice job! Let's fix a few moments to make it look perfect)
Please check yourself with the design
1. Compare the sizes with the design
image
image
2. All prices should be at the same level
image
image
3. Not critical, but the title of the button is different. Also, make sure that all interactive elements have cursor pointer
image
4. Don't forget to compare all the sizes/paddings/font styles with the design
image
image
5. Would be great to navigate the user with links in burger menu
image
6. Also, make sure that your page looks great on different devices using DevTools
image

PR #1105, opened 2024-02-09
Develop

AntonSakh:
-[DEMO LINK] https://AntonSakh.github.io/layout_landing-page/


Moroz-Dmytro (changes requested):
image

3. Why when I click on this text it opens the same page in new tab?(Better just to scroll to the top of page, the same with all other links)

image

4. All addresses should be links to some location on Google Maps. Open them in a new tab with target="_blank".

image

5. There is no padding from the right on smaller screens

image

6. The same on bigger screens, add paddings

image


7. Add paddings from bottom

image

roman-mirzoian (approved):
Well done 👍

PR #1107, opened 2024-02-09
Created The Met landing page

hattoriKanto:
[DEMO LINK](https://hattorikanto.github.io/layout_landing-page/)


polosanya (changes requested):
Almost perfect! 
1. This button does nothing on click. Would be great to navigate user somewhere or open some external page on click
image
2. Make sure that you add transition to the default state of the element, not hover, so it works smoothly both ways
image
3. The font style is wrong here
image
image

roman-mirzoian (approved):
Well done 🔥

PR #1113, opened 2024-02-11
Develop

troleksii:
DEMO LINK - https://troleksii.github.io/layout_landing-page/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add a favicon on the page
image

2. Need to make a call when clicking on this icon
image

image

image

image

6. Add cursor pointer for buttons
image

7. Need to change the autocomplete styles for the inputs
image

8. Add hover effects here
image

l4st1m0za (changes requested):
Good job, but some places can be improved: 

1) This is good idea for the icon hover, but I think that it's better to just increase the icon on the hover
image

2) All the animation should be smooth

3) Photos of bikes should be in the center of their block 
image

4) Make fields required
image

5) Menu and header elements should be placed on the same places. When I open menu, it's elements are not on the same spot

maxim2310 (changes requested):

1. it must looks according design and number must be reachable 
image
image
2. on tablet need to check main img position 
image
image
3. same on mobile
4. check indent here
image

etojeDenys (approved):
lgtm

PR #1118, opened 2024-02-12
add task solution

Dmitrel:
- Deploy and create a Pull Request with a [DEMO LINK](https://Dmitrel.github.io/layout_landing-page/)


BudnikOleksii (approved):
Good work 👍 
Approved, but a few suggestions for the future:
- if you add any validation you should show suggestions what you expect, as for now I can't submit the form
image
- when a user hovers any interactive element, the cursor should become a pointer.

PR #1119, opened 2024-02-12
solution

Yehorf21:
[DEMO LINK](https://Yehorf21.github.io/layout_landing-page/)


anastasiiavorobiova (changes requested):
Great work! A user is not able to scroll the page after clicking on the link in the menu. 
Screenshot 2024-02-12 at 13 18 43
Consider opening phone links in the same tab 
Screenshot 2024-02-12 at 13 24 16
Consider removing the default outlines here. Also, all fields should be required

sTorba24 (changes requested):
Styles didn't apply :( Text your fe_chat to solve issue that you have with it
image

PR #1121, opened 2024-02-12
fixes

Misha408:
[DEMO LINK](https://misha408.github.io/layout_landing-page/)

### DIFF
--- a/src/styles/blocks/menu.scss
+++ b/src/styles/blocks/menu.scss
@@ -72,6 +72,12 @@
       line-height: 27px;
       text-transform: uppercase;
       color: $main-header-font-color;
+
+      transition: color 0.5s;
+
+      &:hover {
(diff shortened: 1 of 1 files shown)


SanyaBratashchuk (changes requested):
1. all external links should be open in the separate tab
2. typo should be message
image
3. check font through the app
4. make all fileds required
5. apply `scroll-behaviour: smooth`

IvanFesenko (changes requested):
Demo link it broken. Please re-check it before request review

2024-02-13_09h56_02

sTorba24 (changes requested):
GJ 🚀 
1. open/close menu buttons should be on the same position to be able to click in same place and having effect close/open

2. add some hover effects on links
image

SanyaBratashchuk (approved):

PR #1126, opened 2024-02-13
add task solution

IvanPovich:
[DEMO LINK](https://IvanPovich.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
image

2. Need to remove scroll when this menu is open
image

3. Add transition for hover effects here and for all buttons
image

4. The user will not be able to change the size of these fields
image

5. These elements should be links and you need to fix font size here as on the design
image

6. Add a placeholder text for the text area and make all these fields are required before sending
image

7. Fix the left indents on the textarea field
image

8. After clicking on the send button, the page should not scroll to the top, but you only need to clear the fields
image

9. Need to fix font size for the text on the mobile version and check it everywhere
image

sTorba24 (changes requested):
Nice progress!

1. add some hover effects
image

2. here too
image


3. do not show phone number on mobile
image

Andriy-Shymkiv (approved):
very good!

PR #1127, opened 2024-02-14
Develop

rodriguesedson:
- [DEMO LINK](https://rodriguesedson.github.io/layout_landing-page/)


src/styles/blocks/contact.css
```diff
@@ -0,0 +1,111 @@
...
+}
+
+.contact__form--button {
+  height: 56px;
+  border-radius: 8px;
+  border: 2px solid #F0F1F4;
+  background-color: #F8F8FA;
+  font-weight: 700;
+  line-height: 27px;
+}
```
EdPirro, line 42: text and button are not modifiers but other elements of the contact block! You could use `contact__text` and `contact__button`

src/styles/blocks/contact.css
```diff
@@ -0,0 +1,111 @@
...
+
+.text--field {
```
EdPirro, line 44: Here, field is not a modifier, maybe you meant to use `-` instead of `--`? Also, you might want to have text-field as an element of the contact block

EdPirro (changes requested):

PR #1128, opened 2024-02-14
Nothing Develop

hatevadym:
[DEMO LINK](https://hatevadym.github.io/layout_landing-page/)


maxim2310 (changes requested):
look like some styles was not connected 
image

etojeDenys (changes requested):
1. better to disable textarea scaling 
image

2. better to import icons as svg
image


3. add the favicon
image


4. looks broken on mobile
image

GUSILLUS (changes requested):
Great job! But

1. You don't need to wrap the image with `<a>` which did nothing
image

2. Add some hover effects here and here
image

image

3. Don't wrap these images with a `<a>` you don't need this in a single-page application it only disappoints users

4. The same goes for the view all button. If you want to make it a link, find a way to show an alert message when the user clicks on a `<a>` only using `href` attribute

5. This needs to be a mail link. Good move with using Nothing link but a better way is to place it on your Nothing logo
image

6. Fix these labels on a sidebar. If you don't have these pages it`s better to show link on your SPA section
image

7. Try to make a better design for these images from 
image
to
image

https://www.awesomescreenshot.com/video/24962901

sTorba24 (changes requested):
1. remove blue background on input autocomplete 
image

2. remove link to `nothing` page from logo
image

sTorba24 (commented):
mentor meant to remove `<a>` wrapper

hatevadym:
зображення

etojeDenys (approved):
lgtm

PR #1129, opened 2024-02-14
add task solution

Ehulinian:
[DEMO LINK](https://Ehulinian.github.io/layout_landing-page/)


maxim2310 (changes requested):
Yor demo link is not working, please fix it and make request review again

l4st1m0za (changes requested):
First end the page to the end, and make in as design in figma. Here are the main diffs:

1) Icons are too small
image

2) Photos should not be overlapping
image

3) This is layout for phones, it shouldn't be applied to the Desctop
image

DarkMistyRoom (changes requested):
- these elements should be bigger. Check in the mockup
image

- show the number on hover on desktop
image

- no need to make 
image

- use the hover icon for the menu opener from the mockup
image

- the Recomended link in the menu should  lead to the appropriate section
image

- fix these images
image
here is how they should look
image

- and add hover effects to them
- fix these elements. How wide the text should be and the indent between the text and the button
image

- please apply the font styles according to the mockup
image

- We don't use this color on the page in other places. Please, use existed color
image

image

SanyaBratashchuk (changes requested):
1. check margins here 
image

SanyaBratashchuk (changes requested):
make sure that you have re-deployed your page since there is still no indents 
image

l4st1m0za (changes requested):
1) Logo seems to be shrinked as in main page, as in menu
image

2) All the animations on the page, should take some time to happen, You have achieved good animation for the menu elements, but for all other clickable elements you should also apply time to work. For all the logos, icons, buttons, etc.

3) Remove the autofill auto styles, and apply your own
[Here is how you can do this](https://css-tricks.com/almanac/selectors/a/autofill/)
image

Viktor-Kost (approved):
### nice!

### to improve:
1. use svg for logo and icons instead of png
2. remove default form submit button border

PR #1131, opened 2024-02-15
correct lint

KatsuboDmytro:
https://KatsuboDmytro.github.io/layout_landing-page/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add a favicon on the page
image

2. Need to fix the background image for the header as on the design
image

3. Need to show the phone number when hovering this icon and make a call after click
image

4. Need to remove scroll when this menu is open
image

5. Add cursor pointer for all clickable elements, example for all buttons
image

6. Fix the left indent for the textarea. Also, the user will not be able to change the size of this field
image

7. After clicking send form I got this problem, need to fix it and make all fields are required before sending
image

8. These elements should be links
image

9. Need to fix the background image position in the footer
image

volodymyr-soltys97 (changes requested):
image
image

2. The same, still not fixed after clicking the form
image

3. These elements should be links and add hover effects here
image

4. The background image in the footer still not fixed, check this image on the design
image

5. Need to check your code with the checklist

IvanFesenko (changes requested):
- Form message are have text  by default, not place holder
https://github.com/mate-academy/layout_landing-page/assets/61904995/6889b65a-579e-4013-81d2-844f7c55d1e6
- Submit still not fixed
2024-02-16_15h46_01

If you have any questions - ask in fe_chat

nazarmatsevych (changes requested):
1. Proofread your code with [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md)

2. Add `scroll-behavior: smooth;` to html selector to make smooth scroll animation
https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior

3. Use mailto for email link, and add a real google map address to address link
image

4. Phone link is not working
image

maxim2310 (changes requested):

1. remove autofill styles 
image
2. address links must be opened in new tabs
3. form must clean up after submit

KatsuboDmytro:
Тільки з JS можна очистити форму, а його я тут поки не використовую

SanyaBratashchuk (changes requested):
2. form should reset on submit. You can implement using following code `const form = document.querySelector(.ClassAppliedForFormElement)  form.addEventListener('submit', (e) => { e.preventDefault(); e.target.reset() })`

nazarmatsevych (approved):
Good work!

Just add a required attribute to form fields and specify the correct input type

PR #1133, opened 2024-02-15
add task solution

ArsenSkandalenko:


volodymyr-soltys97 (changes requested):
Hi, there is no demo link, also you need to add the demo link to the PR description
image

PR #1136, opened 2024-02-17
Develop

MOODDDII:
[DEMO LINK](https://MOODDDII.github.io/layout_landing-page/)


SanyaBratashchuk (changes requested):
1. check paddings 
image
2. make all fields required
3. add scroll for menu if there is no enough space for content

PR #1139, opened 2024-02-19
landing page

Half-a-man:
[DEMO LINK](https://Half-a-man.github.io/layout_landing-page/)


Viktor-Kost (changes requested):
### Looks good!

### To improve:
1. add hover effects to your links
2. add transition to your hover effects (buttons and links) to make it smooth
3. increase transition duration for images. usually we use 250-300 ms
4. The page shouldn't be reloaded on form submit, but the fields should be cleared. Use:
5. add a correct input type to `email` field
6. align `Take the Streets` title at the center
7. footer number link is broken (leads to 404 page). use `tel` at `href`
8. Disable page scrolling under the menu using the next code:
```
}
```
9. value font color and placeholder shouldn't be the same. check figma
Screenshot 2024-02-19 at 21 57 31

maxim2310 (changes requested):

1. would be nice to add hover 
image
2. move this img a bit right, like on design
image

sTorba24 (approved):
Well done!

PR #1151, opened 2024-02-22
add task solution

khromakeiqt:


BudnikOleksii (changes requested):
Add a demo link so a reviewer can test your landing

BudnikOleksii (changes requested):
Good work 👍 
A few fixes are required in accordance with the [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md):
image
image
- In the textarea shouldn't be any prepared text, just placeholder, and outline redundant too.
image
- And in my opinion the custom cursor isn't a good idea for this landing, as it almost match the background color so it dificult to see it.

BudnikOleksii (approved):
Good work 👍 
Approved, just one suggestion, all animations on hover should be smooth with the same timeframe. When I hover links in the menu the border appears immediately

PR #1153, opened 2024-02-23
Develop

NeYaroslav:
[DEMO](https://neyaroslav.github.io/layout_landing-page)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add a favicon on the page
image

2. Need to check the styles for phone number as on the design
image

3. Need to fix position for title of this section as on the design
Demo
image

Design
image

image

5. Need to change the autocomplete styles for form fields. Also, after clicking send, the page should not reload
image

6. Make resize none for the textarea
image

7. Need to hide the phone number on the mobile and tablet versions
image

Esceype (approved):
Well done!

PR #1154, opened 2024-02-23
add task solution

VladKugot:
[DEMO LINK](https://VladKugot.github.io/layout_landing-page/)


Esceype (changes requested):
GJ!
Need to fix:
- Here should be a hover effect as in design:
image
image

- This text should be alike as it is in the menu and start a call
image

The submit button doesn`t work
- These images should have hover effects as the images below
image
- Address in the contact us section should also be a link(Checklist point 6 - https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md)
- Disable scroll when menu is open
image
- All buttons should have cursor pointer and hover effects, same about images/icons

anastasiiavorobiova (changes requested):
Great work! 
Screenshot 2024-02-25 at 15 48 41
Consider removing outlines, and default borders and checking the font here 
Screenshot 2024-02-25 at 15 47 58
Check margins here

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add a favicon on the page
image

2. Need to change the autocomplete styles
image

3. Add hover effects here
image

4. This link must be open in the new tab
image

5. Need to reset form field after send form, also make all these fields are required before sending
image

6. Need to hide phone number on the mobile and tablet versions
image

7. Need to fix the background image here as on the design
image

anastasiiavorobiova (changes requested):
Great work! Almost done! 
Screenshot 2024-02-26 at 15 03 39

GUSILLUS (changes requested):
Good job! But

1. You need to add some action on a button. Maybe some alert message
https://www.w3schools.com/tags/att_a_href.asp

2. You need to fix your submit button it only needs to reset the form. For now, it is redirection to this page
image

3. Make this link go to the section inside your landing like Recommend, Category, About us, Contact Us
image

4, Fix the image on the 320px breakpoint from 
image
to this 
image


If you have some questions you can ask them in a fe_chat or q&a

PR #1156, opened 2024-02-23
Develop

YuliiaCat:
https://yuliiacat.github.io/layout_landing-page/


vadiimvooo (changes requested):
1. You need to add transition on hover effect for this link:
image
2. Remove scroll when burger menu is opened.
image
3. It will be good if you add some interaction on photos in sections (maybe some scale)
4. Wrong paddings for input and redundant scale bar for textarea
image
5. When I start typing in text area placeholder don't dissapear, please fix it.
image

nazarmatsevych (approved):
Great work!

You can also add some hover effects and cursor: pointer to all interactive elements
image
image

PR #1157, opened 2024-02-25
landing page bem

NKoshmak:
[DEMO LINK] https://NKoshmak.github.io/layout_landing-page/


l4st1m0za (changes requested):
Good start! Now you need to finish all lending and then call for another pull request!

PR #1159, opened 2024-02-26
some code to solve mistakes

Leh-02:
There are some problems . On the desktop (and other devices), the scroll appears when the screen width is reduced.  I can't figure out how to remove it, so that firstly the pictures in the grid are as in the design and the text is transferred


volodymyr-soltys97 (changes requested):
If you need a help, feel free in the fe_chat, now you need to fix the horizontal scroll

GUSILLUS (changes requested):
GJ!

1. Add max-width propertry to improve ui on a big screen
2. User need ability to click on it
image

3. Consider adding some hover effects here
image

4. Same
image

5. Same 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add a favicon on the page 
image

2. Fix font weight for the prices 
image

image

4. Fix height for the inputs, check it on the design 
image

image

6. This link must be open in the new tab 
image

7. Hide the additional phone number on the mobile and tablet versions 
image

anastasiiavorobiova (changes requested):
Great work! 
Screenshot 2024-09-24 at 22 27 33
The menu should be scrollable in landscape mode 
Screenshot 2024-09-24 at 22 26 53
Check fonts here

volodymyr-soltys97 (approved):
Great work! 🔥

PR #1162, opened 2024-02-27
The MET page

IrynaZahorodnia:
[DEMO LINK](https://IrynaZahorodnia.github.io/layout_landing-page/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -171,6 +171,34 @@
         "jsesc": "^2.5.1"
       }
     },
+    "@babel/helper-annotate-as-pure": {
+      "version": "7.22.5",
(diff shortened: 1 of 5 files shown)


vadiimvooo (changes requested):
1. Add transition effect for this links as well
image
2. Scale should not overlap the neighbor elements.
image
3. Add transition effect for the button
image
4. The same is for this links
image

GUSILLUS (changes requested):
GJ! But:

1. Add some alert logic to all buttons instead of scrolling to top
image
Look [here ](https://www.w3schools.com/tags/att_a_href.asp#:~:text=Definition%20and%20Usage,top%20of%20the%20current%20page!)how to do it

2. Make all images just a standard image with a hover effect in your case link wrap is redundant
image

3. Fix header block image on all breakpoints to make it more similar to design

nazarmatsevych (approved):
Great work 🔥

PR #1163, opened 2024-02-27
add solution

AndriiZaharuk:
- Deploy and create a Pull Request with a [DEMO LINK](https://AndriiZaharuk.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good start 👍 
Let's improve your page
1. Add a favicon on the page
image

2. Add the left and right padding for the content. Check it on the design
image

image

4. Add transition when this menu appears on the page
image

5. Add hover effects for all clickable elements on the page

Keep going with the next sections

vadiimvooo (changes requested):
Good job, you are doing well. Just one moment. You need to change hover effect for burger menu links, it should look like on the photo. Also text for this links has to be in capitalized case instead of upper case. Also provide a transition.
image

l4st1m0za (changes requested):
Good job, now you need to end the landing to the end, and ask for the pr review when it will be done completely. 

Also, you need to change this hover effect:
image

When user hovers over the element, it should be only underscored, without changing the case.

l4st1m0za (changes requested):
Good job for now, the only thing, you need to add smooth transition for the elements in the menu. Now, call for the next review, after ending whole landing

lerastarynets (changes requested):
GJ!
pls implement the whole landing and rerequest the review

PR #1170, opened 2024-02-29
add task solution

silchenko-arsen:


maxim2310 (changes requested):
Plz add demo link to PR

PR #1171, opened 2024-02-29
Bikes_Landing

dmytro-prontenko:
Please check


maxim2310 (changes requested):

1. font is different from design 
image
2. would be much better if img don't overlap near blocks 
image
3. remove autofill styles 
image
4. you have 3 different fonts here 
image
5. fix repeat img 
image
6. it must be aligned, and would be nice to add hover effect on links
image

volodymyr-soltys97 (changes requested):
To improve: 
1. Why did you remove the autocomplete? This is wrong solution, you need to change the autocomplete styles and you need to enable autocomplete
image

2. Add transition for hover effects everywhere
image
image

image

4. Make these fields are required before sending
image

5. Need to fix styles for inputs text as on the design, check it. Also need to remove default border when input is focusing
image

6. Need to make a call when clicking on this icon
image

dmytro-prontenko:
fixed

IvanFesenko (changes requested):
Good job: Check commetns should be fixed:
- Header section should have 100vh height
2024-03-01_15h30_10
- All fields in form should be required
- Menu items jumps on hover
https://github.com/mate-academy/layout_landing-page/assets/61904995/b594d2d1-0bd0-4592-9e92-af5b4364c012
- Add transition for all hover effects

PR #1174, opened 2024-03-01
add task solution

silchenko-arsen:
[DEMO LINK](https://%3Cyour_account%3E.github.io/layout_landing-page/)


l4st1m0za (changes requested):
Your demo liknk is broken, try to redeploy your project!

l4st1m0za (changes requested):
1) Main photo is strangerly cropped, also use .svg for the icons
image

2) Menu is broken, also add smooth scroll for scrolling to the needed places on the site with the help of menu links
image

3) Elements in the menu, should not scale on hover, but should be underlined on the hover

4) Photos here should scale on the hover
image

5) Remove autofill styled from the email input https://css-tricks.com/almanac/selectors/a/autofill/

6) Icons should be of the same size
image

l4st1m0za (approved):
Good job!

PR #1179, opened 2024-03-03
MyBike

nester7an:
-  [DEMO LINK](https://nester7an.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Need to show phone number when hovering this icon, check it on the design
image

2. The user will not be able to change the size of these fields
image

3. Add cursor pointer for this button and remove the default border
image

4. Need to change the autocomplete styles
image

5. Need to fix indents between these elements, check it on the design. Also add transition for hover effects here
image

maxim2310 (changes requested):

1. i must be able click on it 
image
2. email must trigger create mail message
image
3. need remove autofill styles 
image
4. also check horizontal scroll 
image

maxim2310 (changes requested):
not fixed 
image

l4st1m0za (approved):
Good job!

PR #1183, opened 2024-03-04
add task solution

vladkoval05:
[DEMO LINK](https://vladkoval05.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
image

2. Need to fix the background color here, he must be the same everywhere
image

3. Need to add hover effects for all clickable elements on the page
image

4. Add some hover effects for the images
image

5. The font styles for the placeholder are different here, so you need to fix it
image

6. The user can not be able to change size of this field
image

7. Need to add cursor pointer for this button and add transition for hover effects
image

8. Need to change the autocomplete styles for inputs, and need to remove default border when input is focusing
image

9. These elements must be links and add hover effects here
image

10. Need to fix the background image position on the footer, check it on the design
image

anastasiiavorobiova (changes requested):
Great work! 
Screenshot 2024-03-10 at 18 29 18
The mobile layout looks broken 
Screenshot 2024-03-10 at 18 29 00
Consider exporting an image without a label here 
Screenshot 2024-03-10 at 18 28 43
Consider adding proper href attributes here

loralevitska (approved):
Good job 🔥

PR #1190, opened 2024-03-08
add task solution

jkopan:
[demo link](https://jkopan.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add a favicon on the page and remove 'landing' word from the title of the page
image

2. The left and right indents are different, so you need to fix it
image

3. Remove position for the footer
image

4. The user cannot be able to change size for the textarea
image

5. Need to fix the left indent for the text and remove default border when input is focusing
image

6. Need to show phone number when hovering this icon
image

7. The background color must be is same, need to fix it
image

sTorba24 (changes requested):
Nice progress!

1. images shouldn't overlap each other
image

2. add hover effects on buttons(check mockup)
image
image

3. link that trigger mail and cursor pointer should be
image


4. check mobile version, it has weird right margin
image


5. icons should be on the same lvl
image

maxim2310 (changes requested):

1. here differents font family 
image
2. for centering 
image

loralevitska (changes requested):
You did an amazing job! Looks good! Let's do some improvements (check comments)

- center phone number
image
- don't show phone number on mobile mode
image
- fix menu according mockup 
image

sTorba24 (approved):
Good work! 

1. trigger mailing
image

PR #1191, opened 2024-03-10
landing B&O

ElleeD:
[DEMO LINK](https://ElleeD.github.io/layout_landing-page/)


maxim2310 (changes requested):

1. here must be a phone on hover icon and it must be clickable 
image
2. check it according design 
image
3. remove quotes 
image
4. text area must be not resizable and have placeholder 
image
5. need add validation to inputs 
6. remove autofeel styles
image

DarkMistyRoom (approved):
Good job! Here are a few additional comments:
- please, make the number reachable by the cursor
image

- don't use blue color for the hover effect on these links. We don't have this color anywhere on the page. Use some color from the mockup
image

image

PR #1200, opened 2024-03-13
add layout

lerika7:
[Nothing](https://www.figma.com/file/DtkQmQ797hk0nI4KfMi2Uq/BOSE-New-Version?type=design&node-id=6802-139&t=L7eKz5YKLN0m5WxR-0)


volodymyr-soltys97 (changes requested):
Hi, there is no demo link, you need to add the demo link here and add this link to the PR description

IvanFesenko (changes requested):
Good job. Check my comments:
- all fieldls in form should be required, remove autocomplete styles, form submit should clear form, text should starts on to in message filed, not center
2024-03-14_16h45_47
- should be links as well
2024-03-14_16h47_21
- horizontal scroll on mobile resolution

PR #1202, opened 2024-03-14
add solution

ArtemZakhar:
[DEMO LINK] (https://ArtemZakhar.github.io/layout_landing-page/)


src/styles/blocks/communication.scss
```diff
@@ -0,0 +1,19 @@
...
+    font-weight: 300;
+  }
+
+  &__information {
+    margin-bottom: 24px;
+    font-size: 16px;
+    line-height: 140%;
+    font-weight: 400;
+
+    &:last-child {
```
mrkovaliv, line 15: better use

```
&__information:not(:last-child) {
  margin-bottom: 24px;
}
```

mrkovaliv: also better use this approach to improve readability 

 `&:last-child {` ->  `&__information:last-child {`

mrkovaliv (changes requested):
Great job 🚀 🚀 🚀 

Added some small comments))

src/styles/blocks/communication.scss
ArtemZakhar: @mrkovaliv, look, I had a problem with the sticky menu. If I throw away the display none property - it stays on the screen and takes 100vh place(
I don't understand why it happening because in the lesson there was shown a possibility to exclude display manipulations.

mrkovaliv: Could do you please ask this question in q&a chat (I think it'll be useful for other students)?

PR #1207, opened 2024-03-16
try number 1

igols:
https://igols.github.io/layout_landing-page/


vadiimvooo (changes requested):
1. This text should appear only when you hover phone icon.
image
2. Remove scroll when you open menu
image
3. Something wrong with your footer, please fix it
image
4. This section should be fixed, please check the mock up
image
5. You don't need to move section title (you haven't to use position sticky for it)
image
6. Contact us seciton should also be fixed, please check the mockup
image

volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. Add a favicon on the page
image

2. You need to show these phones number only when hovering the phone icon
image

image

image
image

5. You need to fix height for the inputs, also need to change the autocomplete styles
image

maxim2310 (changes requested):

1. where did phone icon go?)
image
2. is it feature, when i click on logo the main text change color?
image
3. not fixed
image
4. still no hover effect here
image
image
image
5. also not fixed 
image
6. fix indent between img, check design 
image
7. and check mobile view 
image

PR #1214, opened 2024-03-17
Task implementation

kinqbert:
[DEMO LINK](https://kinqbert.github.io/layout_landing-page/)

p.s. I've added some margins to separate blocks of the website, this is not according to BEM rules, but still these blocks won't be used anywhere else and also code will be looking much more messier with additional containers added just for margins.


DanilWeda (changes requested):
Good job. 
Check some UI improvements: 

1. Add smooth for number ( when it appears)
image

2. Make action for photo in Details section ( transition, cursor)
image

3.  Work with sticky footer. ( You need add image as background for this effect)  
image

Many thanks)

DanilWeda (approved):
Great. 
Approved.
Many thanks)

PR #1216, opened 2024-03-17
add solution

NKoshmak:
[DEMO LINK](https://NKoshmak.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add a favicon on the page
image

2. It is necessary to fix the styles for the phone number as in the design, also the number now disappears when you hover over it, this needs to be fixed
image

3. The user cannot be able to change size of this field
image

image

5. The width for the inputs and button must be are same.
Also make all these fields are required before sending
image

6. The images should not overlap each other when hovering
image

7. You need to fix the horizontal scroll on the mobile version
image

8. Need to hide the phone number on the mobile and tablet versions
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
image

2. Add hover effects here 
image

etojeDenys (approved):
it's better to fix these small jumps
https://github.com/user-attachments/assets/40d39493-794e-42df-b39e-98b42ebf1f4b

but looks good in general

PR #1219, opened 2024-03-18
Develop

DaniilManakov:
[DEMO LINK](https://daniilmanakov.github.io/layout_landing-page/)


l4st1m0za (changes requested):
Good job, but some improvements needed:

1) All the interactive elements should have hover effect (icons, photos, buttons, links, etc)
image

2) All the animation should be smooth and silky

3) Remove scroll in the menu

4) All these links should be working. Email link should open email, and address, should open maps with needed location in it.
image

5) Check design for this section
image

6) Remove basic outline for the form inputs. Also remove autofill styles, here is the link which will help you https://css-tricks.com/almanac/selectors/a/autofill/
image

7) For should be clear after the submitting

PR #1220, opened 2024-03-18
The MET landing page

odunchyk:
[DEMO LINK](https://odunchyk.github.io/layout_landing-page/)


src/index.html
```diff
@@ -1,22 +1,526 @@
...
+            src="images/THE MET.png"
+            alt="The MET logo"
+            class="top-bar__logo"
+          />
+        </a>
+
+        <div class="top-bar__icons">
+          <a
+            href="#contact-us"
+            class="icon icon--phone"
```
mrkovaliv, line 55: Perhaps, it'll be better `icon` -> top`-`bar__icon (But it isn't blocker)

src/styles/blocks/menu.scss
```diff
@@ -0,0 +1,51 @@
...
+  font-weight: 700;
+  color: $main-white;
+  background-color: $accent-main;
+
+  overflow: auto;
+
+  &__top {
+    margin-bottom: 40px;
+
+    position: sticky; /* to keep cross always visible */
```
mrkovaliv, line 13: remove all unnecessary comments

src/styles/blocks/menu.scss
```diff
@@ -0,0 +1,51 @@
+.menu {
+  height: 100vh;
```
mrkovaliv, line 2: better `min-height`

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,27 @@
+.header {
+  height: 100vh;
```
mrkovaliv, line 2: better `min-height`

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,27 @@
...
+
+  background-image: url(../images/header-bg.png);
```
mrkovaliv, line 4: you could write the same but shorter: `background: ...`

src/styles/blocks/contact-us.scss
```diff
@@ -0,0 +1,165 @@
...
+      line-height: 48px;
+
+      grid-column: 2 / 6;
+    }
+
+    @include on-desktop {
+      grid-column: 5 / 9;
+    }
+
+    &-wrapper {
```
mrkovaliv, line 24: `contact-us__title-wrapper` is the other element so don't insert into `contact-us__title`

src/styles/blocks/explore-online.scss
```diff
@@ -0,0 +1,80 @@
+.explore-online {
+  &__title {
+    &-wrapper {
```
mrkovaliv, line 3: the same BEM problem

src/styles/blocks/now-on-view.scss
```diff
@@ -0,0 +1,158 @@
...
+      @include on-desktop {
+        grid-column: span 4;
+      }
+    }
+  }
+
+  &__expo {
+    margin-bottom: 48px;
+
+    &-text {
```
mrkovaliv, line 116: the same BEM problem

src/styles/blocks/now-on-view.scss
```diff
@@ -0,0 +1,158 @@
...
+      font-size: 16px;
+      font-weight: 400;
+      line-height: 27px;
+
+      &-title {
```
mrkovaliv, line 121: the same BEM problem

src/styles/blocks/menu.scss
```diff
@@ -0,0 +1,51 @@
...
+    flex-direction: column;
+
+    gap: 8px;
+    margin-bottom: 16px;
+  }
+
+  &__phone-number {
+    font-size: 16px;
+    line-height: 27px;
+    color: inherit;
```
mrkovaliv, line 38: you don't need this style

src/styles/blocks/menu.scss
```diff
@@ -0,0 +1,51 @@
...
+
+  &__call-to-order {
+    width: fit-content;
+    padding-bottom: 7px;
+    border-bottom: 1px solid $main-white;
+
+    font-size: 12px;
+    line-height: 16px;
+    letter-spacing: 2px;
+    color: inherit;
```
mrkovaliv, line 49: you don't need this style

mrkovaliv (changes requested):
Great job 👍 👍 👍 

Added some small coments)))

src/styles/blocks/menu.scss
odunchyk: Can't understand why color: inherit there is redundant.
It is class used in <a> tag, so without this style my color on links in menu reverts back to standart blue of <a> tag.
In new commit i moved color: inherit to style <a> tag, so all links will inherit father color, but i don't really sure if it's a right solution.
Can you explain a llitle bit further why i don't need this style?

DanilWeda: Using `inherit` is not a good idea, because we can get a style side effects. 
Our color is based on the parent element. If in the future change the color in a parent element, also, changes the color child element. ( this is a side effect). 
In this case, it is better to use those colors that should definitely be here, so that their value needs to be changed additionally and the developer does not miss this moment and does not allow a side effect.

DanilWeda (changes requested):
Good job. 
Check my answer about inherit ( and check, maybe u have cases where you do not need to change color in general.)
Also, add the phone (on the left side) when the user hovers the phone icon.
image
Many thanks)

alexander-ignatow (approved):
LGTM! 👍 

One minor thing (for the future)

This one should probably be a `tel:` link
image

PR #1232, opened 2024-03-20
test

igols:
https://igols.github.io/layout_landing-page/#


GUSILLUS (changes requested):
GJ! But: 

1. Fix the design on a big screen
image

2.  Try to change the hover effect
image

3. Same 
image

4. Make the sidebar more similar to the design


5, Make responsive design more similar to the design on every breakpoint

GUSILLUS (changes requested):
GJ! Try to add max-width on a landing content

1. Fix hover effect in side bar
image

2. Add ability to click on this link
image

3. Add some [alert logic](https://www.w3schools.com/tags/att_a_href.asp) on this button instead of nothing
image

4.Fix this block in 740px breakpoints
image

VitaliyBondarenko1982 (changes requested):
Good job.
Needs some imprvements.
- Still not fixed mismatches wiht design.
actual:
chrome-capture-2024-3-21 (2)
design: 
Screenshot from 2024-03-21 18-08-54
- remove default margins for body
Screenshot from 2024-03-21 18-02-40
- can't to click on phone number in header
chrome-capture-2024-3-21 (1)
- fix horizontal scrolls
Screenshot from 2024-03-21 18-10-40
- transition should work in two ways. for it set transiion on selector (not on selector:hover)
chrome-capture-2024-3-21 (2)


If you have some questions - feel free to ask in fe chat

SanyaBratashchuk (changes requested):
1. apply `cursor: pointer` for all interactable elements 
2. check paddings for input's value
image

SanyaBratashchuk (approved):
add scroll for menu if there is no enough height to show content

PR #1245, opened 2024-03-22
add solution

D-Kondratiev:
Deploy and create a Pull Request with a [DEMO LINK](https://D-Kondratiev.github.io/layout_landing-page/)


src/index.html
```diff
@@ -1,22 +1,315 @@
...
+              >
+                <div class="form__inputs">
+                  <input
+                    type="text"
+                    minlength="3"
+                    maxlength="50"
+                    name="name"
+                    class="form__input"
+                    placeholder="Name"
+                    pattern="[A-Z][a-z]+"
```
undeletable, line 260: [Not critical]
I don't think we need to limit minimal and maximal length for a name, as well as require it to match pattern.
Usually, if there's such kind of restrictions, user should be explicitly informed about that: with placeholder, tooltip, help text etc.
This is not critical for a current task, but it'd be nice to mind that during the future work.

src/styles/blocks/icon.scss
```diff
@@ -0,0 +1,31 @@
...
+    height: 32px;
+    width: 32px;
+  }
+
+  &--phone {
+    background-image: url(/src/images/icons/Icon-Phone-call.svg);
+  }
+
+  &--menu {
+    background-image: url(/src/images/icons/Icon-Burger-menu.svg);
```
undeletable, line 19: [Not critical]
The approach with having separate images for regular and hover states works fine. But since the difference in their sizes is not big, I'd better consider using the same image, but with different width/height for these states.

src/images/backround-img/header-bg-mini.png
undeletable, line 1: [Not critical]
This image is not used and can be safely removed

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,42 @@
+.header {
+  background-image: url(/src/images/backround-img/header-bg-mobile.png);
```
undeletable, line 2: [Not critical]
The header image for narrow screens doesn't look well enough - there's too much empty space of the image displayed:
header320
I'd suggest to play with media query+background styles to improve this.

undeletable (changes requested):
This is really nice work. The page looks well and smooth, and the code is well-organized.

The only major thing to improve is its look for narrow mobile devices. The contact form is not centered in this case:
form320

I've also added some other notes, marked as [Not critical]. These are rather advices for your further work. If you leave these places unchanged, this would not be a reason for rejecting the solution.

src/styles/utils/var.scss
```diff
@@ -0,0 +1,11 @@
+$tablet-min-width: 576px;
+$desktop-min-width: 1024px;
+$main-text-color: #fff;
+$menu-text-color: #1d1d1d;
+$input-placeholder-color: #7e7e83;
+$input-background-color: #f8f8f8;
+$main-background-color: #1d1d1d;
+$menu-background-color: #c2c2c2;
+$effect-duration: 0.3s;
+
```
undeletable, line 10:
```suggestion
$padding-inline: 20px;
$padding-inline-tablet: 72px;
$padding-inline-desktop: 120px;
```

src/styles/utils/mixins.scss
```diff
@@ -0,0 +1,54 @@
...
+}
+
+@mixin on-desktop {
+  @media (min-width: $desktop-min-width) {
+    @content;
+  }
+}
+
+@mixin content-padding-inline {
+  padding-inline: 20px;
```
undeletable, line 14:
```suggestion
  padding-inline: $padding-inline;
```

src/styles/utils/mixins.scss
```diff
@@ -0,0 +1,54 @@
...
+
+  @include on-tablet {
+    padding-inline: 72px;
+  }
+
+  @include on-desktop {
+    max-width: 1020px;
+    margin-inline: auto;
+    padding-inline: 120px;
```
undeletable, line 23:
```suggestion
    padding-inline: $padding-inline-desktop;
```

src/styles/blocks/form.scss
```diff
@@ -0,0 +1,44 @@
...
+  &__input {
+    box-sizing: border-box;
+    border: 1px solid transparent;
+    padding: 13px 15px;
+    border-radius: 40px;
+    color: #292929;
+    background-color: $input-background-color;
+    font-size: 14px;
+    font-weight: 400;
+    line-height: 20px;
```
undeletable, line 17:
```suggestion
    max-width: calc(100vw - 2 * $padding-inline);
    
    @include on-tablet {
      max-width: none;
    }
    
    @include on-desktop {
      max-width: none;
    }
```

undeletable:
@D-Kondratiev 
I've noticed that inputs behave in an odd manner when paddings are set for them. I couldn't find any explanation of such browser's behavior, but there's a way around it. I set additional style fo `.form__input`:
```
max-width: calc(100vw - 40px);
```
where 40px is double the value of `padding-inline` for `.container`. This allows to make sure that input is no wider than page width minus paddings, so it will never be wider than we want.
I've added code suggestions for the way how this style could be added. You can try them out.

Since this is non-obvious thing, I'm approving you solution. Just mind these input issues during the future work.

undeletable (approved):
Well done! Please check the comments.

PR #1251, opened 2024-03-23
solution

Mykhalishere:
- Deploy and create a Pull Request with a [DEMO LINK](https://Mykhalishere.github.io/layout_landing-page/)


Viktor-Kost (approved):
### Good job!

### Recommendation
Usually we don't use such hover effects as you've applied here (looks like a glitch). Make it simpler
Screenshot 2024-03-23 at 14 18 42
Screenshot 2024-03-23 at 14 18 42

PR #1254, opened 2024-03-24
done lending

yanamomot:
DEMO: https://yanamomot.github.io/layout_landing-page/


src/header/header.scss
```diff
@@ -0,0 +1,79 @@
+.header {
+  box-sizing: border-box;
```
alexander-ignatow, line 2: I advice you to have some common `box-sizing` value, set in the reset file

src/index.html
```diff
@@ -11,12 +11,231 @@
...
+           placeholder="Name">
+
+          <input class="contacts__email" type="email"
+          name="email"
+           placeholder="Email">
+
+          <textarea class="contacts__message" name="message"
+          cols="30"
+           rows="10" placeholder="Message"></textarea>
+          <a class="link contacts__send" href="#">Send</a>
```
alexander-ignatow, line 223: Please, wrap these with the form and use the button (or input with `type=submit`) for the `Send`

alexander-ignatow (changes requested):
First of all - good job 👍 
I have some comments mainly on the file structure

- Please, reorganize your style files, according to the BEM structure: you need to have one file per one BEM block
- In your case, you can set font-family in reset/normalize and get rid of `fonts.scss` (same for is true for `box-sizing`)

src/fonts-reset-var/normalize.scss
```diff
@@ -0,0 +1,176 @@
...
+}
+
+h2{
+font-family: 'Space Grotesk', serif;
+font-weight: 700;
+font-size: 32px;
+line-height: 32px;
+}
+
+.header__title{
```
undeletable, line 13: Starting from this line, styles in `normalize.scss` correspond to different BEM blocks and are not related to normalization - so they have to be moved to separate file(s).

PR #1258, opened 2024-03-25
Develop

OTsiolkovskiy:
[DEMO LINK](https://OTsiolkovskiy.github.io/layout_landing-page/)


src/styles/blocks/compare-bikes.scss
```diff
@@ -0,0 +1,65 @@
+.compare-bikes {
+  margin-bottom: 92px;
```
alexander-ignatow, line 2: No margins on BEM blocks please

src/styles/blocks/details.scss
```diff
@@ -0,0 +1,87 @@
+.details {
+  margin-bottom: 80px;
```
alexander-ignatow, line 2: Margin on BEM block

src/styles/blocks/icons.scss
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  &__menu {
+    @include square($-side-of-icon);
+
+    margin-left: 30px;
+  }
+}
+
+.icon {
```
alexander-ignatow, line 50: This is different BEM block - pleas put it into different file

alexander-ignatow (changes requested):
Looks good 👍 , but some improvements are needed

1. Hero image positioning looks wrong:
image

While you have :
image

2. Please re-check these margins
image

3. Why do you need the `fonts` folder with the font file?

4. Some other comments are in the code

src/styles/blocks/compare-bikes.scss
OTsiolkovskiy: 1.  When the screen size matches the size specified in the figma layout, hero image positioning looks good. But when the screen size changes, in particular, the height decreases relative to the width, hero image positioning looks different. To fix this, I tried changing the values of the background-size and background-position properties, as well as changing the height relative to the width. However, the result was worse positioning of the background-image. That's why I didn't make any changes to the code for this comment.
2. Checked and fixed.
3. Fixed.

src/styles/blocks/details.scss
OTsiolkovskiy: fixed

src/styles/blocks/icons.scss
OTsiolkovskiy: fixed

alexander-ignatow (approved):
Good! 👍

PR #1267, opened 2024-03-27
Develop

tarassamchenko:
https://tarassamchenko.github.io/layout_landing-page/


polosanya (changes requested):
Nice start! Keep up the good work!
1. Check yourself with the design. The logo and icons are too big
image
2. The phone number should appear here on hover
image
3. Burger menu doesn't match the design either. Also, there should be no scroll in it
image
4. Font styles of the prices are wrong
image
5. This button makes nothing on click. Also, hover effect doesn't work smoothly
image
6. Address link should open a Google Maps page in a new tab. Also, all links should have some hover effect
image
7. Form size doesn't match the design
8. Paddings should be the same for all sections. Same with tablet version.
image

PR #1271, opened 2024-03-28
MyBIKE lending solution

ilyasolo881:
- Deploy and create a Pull Request with a [DEMO LINK](https://ilyasolo881.github.io/layout_landing-page/)

Not all checklist check's was completed, but look's good :)


src/styles/blocks/_contacts.scss
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  &__button {
+    margin-bottom: 48px;
+    width: 100%;
+    transition: box-shadow 0.3s;
+  }
+}
+
+.form {
```
alexander-ignatow, line 79: New BEM block in the should be in a separate file

src/styles/blocks/_contacts.scss
```diff
@@ -0,0 +1,140 @@
...
+
+textarea:hover {
+  border: 1px solid $color-input-hover;
+  color: $color-input-hover;
+}
+
+textarea:focus {
+  border: 1px solid $color-input-active;
+  color: $color-input-active;
+}
```
alexander-ignatow, line 140: Please style this based on classes and move to the separate files

src/styles/blocks/_section-title.scss
```diff
@@ -0,0 +1,16 @@
+.section-title {
+  font-weight: 500;
+  font-size: 32px;
+  line-height: 100%;
+  letter-spacing: 1.5px;
+  text-align: center;
+  color: #fff;
```
alexander-ignatow, line 7: Maybe this should be in variables?

alexander-ignatow (changes requested):
Good job! 
Let's improve your solution a bit:

1. Please fix how the hover effect works on the phone number in the header
image

image

2. It would be nice to have enter / exit animation on the site menu

3. Please fix some minor issues through the code

ilyasolo881:
Fix's:
1. Variable in _section-title.scss
2. Separate and add class to textarea and input
3. Move block form to another file
4. Add transition to menu 
Not fixed:
1. Can't understand how to animate phone nubmer in top-bar.
try to add father class to phone and number to animte his hover
try to set on phone:hover opacity number
try use nesting in phone:hover (+ or ~) number 

but not work's :(

src/styles/blocks/_form.scss
```diff
@@ -0,0 +1,42 @@
+.form {
```
undeletable, line 1: As far as I see, `form` classname is not used at all, so this file is not needed anymore

src/styles/utils/_normalize.scss
```diff
@@ -0,0 +1,10 @@
+* {
```
undeletable, line 1: Ideally, we should not use `*` selector, as it might have negative impact on performance

undeletable:
@ilyasolo881 
> Can't understand how to animate phone nubmer in top-bar.

Try using pseudoelements (`before`/`after`) for phone number

src/styles/blocks/_contacts.scss
```diff
@@ -0,0 +1,101 @@
...
+    &:focus {
+      border: 1px solid $color-input-active;
+      color: $color-input-active;
+    }
+  }
+
+  &__textarea {
+    width: 100%;
+    min-width: 288px;
+    max-width: 405px;
```
undeletable, line 74: Setting different max widths for different form fields (textarea, inputs) causes them to have different sizes on some window widths:
form375
I'd recommend setting `max-width: 100%;` for all form fields at least for small screens, in order to avoid layout issues.

src/index.html
```diff
@@ -1,22 +1,409 @@
...
+
+              <p class="details__card-text">
+                The app senses when you're nearby to unlock automatically. GPS
+                tracking so you know where your bike is and can track it
+                anytime.
+              </p>
+            </article>
+
+            <article class="details__card">
+              <div class="details__photos">
```
undeletable, line 291: According to the design, pictures should go in different order here

undeletable (approved):
Despite some inaccuracy, this is a good work, so I'm approving it. Please take a look at the comments and consider them in your future work.

PR #1272, opened 2024-03-28
add task solution

HannaMatviienko:
[DEMO LINK](https://hannamatviienko.github.io/layout_landing-page/)


SanyaBratashchuk (changes requested):
1. apply `cursor: pointer` for all interactable elements 
2. it is better to use svg images for icons/logos

SanyaBratashchuk (approved):

PR #1279, opened 2024-04-03
landing

BHIgor:
[DEMO LINK](https://BHIgor.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. should be the links
image

2. Use the logo as a favicon
image

3. 
image

4. 
image

5. add to buttons `cursor: pointer`
6. 
image

src/index.html
```diff
@@ -9,14 +9,369 @@
...
+              required
+              >
+
+              <textarea
+              type="text"
+              class="main__inpt--area"
+              placeholder="Messaege"
+              required
+              >
+              </textarea>
```
etojeDenys, line 322:
```suggestion
              <textarea
              type="text"
              class="main__inpt--area"
              placeholder="Messaege"
              required
              ></textarea>
```
to fix spaces

etojeDenys (changes requested):
1. set height: 100vh to the header on desktop
image
looks better with that


2. should start the call on click
image
image

3. fix the default spaces in textarea
image

4. should lead to some section in the app
image

volodymyr-soltys97 (changes requested):
Good job!
Let's improve your page
image

2. Now it is not possible to hover over the phone number, it immediately disappears, we need to fix it so that it is possible to hover over the phone number
image

image

4. You need to remove default border when input is focusing
image

5. You need to remove default border for the input and button
image

6. You need to fix the background image position, check it on the design
image
image

l4st1m0za (changes requested):
1) Add hover on all the interactive elements(logo, icons)
2) Remove the autofill styels from the email input, https://css-tricks.com/almanac/selectors/a/autofill/

l4st1m0za (approved):
Set your favicon so that it lookss good!

PR #1282, opened 2024-04-03
added solution

yuriipodria:
- [Demo link](https://yuriipodria.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. You need to show the phone number when hovering this icon and make a call when clicking here
image

2. The price in all blocks must be horizontally aligned
image

3. The use cannot be able to change size of this field
image

4. You need to remove the underscores in these elements 
image

5. Add actual links for these elements, you need to make a call when clicking on this icon, open the mail client, аnd open the google maps
image

volodymyr-soltys97 (changes requested):
GJ!
image

2. Add actual address on google maps here
image

image

GUSILLUS (changes requested):
GJ! But:

1. You can add some [alert logic](https://www.w3schools.com/tags/att_a_href.asp) instead of link #
image

2. Disable scroll in menu
image

3. Remove this from menu. You don't have it on design
image

vadiimvooo (changes requested):
Good job. One thing to fix:
1. Hover of the contact us section elements should not move the neighboors
image

Viktor-Kost (approved):
### Nice!

### Recommendations:
1. use svg format for a logo image instead of png one

PR #1284, opened 2024-04-04
Develop

aPetrashyshyn:
[DEMO LINK](https://aPetrashyshyn.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job!
Let's improve your page
1. When you hover over the phone number, it disappears, you need to fix it so that you can hover over the phone number
image

2. The images should not overlap each other when hovering
image

3. The user cannot be able to change size of this field
image

4. Remove default border when input is focusing
image

5. Need to remove the default border for the button
image

SanyaBratashchuk (changes requested):
1. all fields should have the same font-family
image
2. disable page scroll when menu is open 
3. check font-color 
image

PR #1286, opened 2024-04-04
add task solution

raccoon1Andrew:
[DEMO LINK](https://raccoon1Andrew.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. let's fix this overlapping
image

2. number should appear on hover
3. you should start the call on click
image

l4st1m0za (changes requested):
1) On hover on this ellement, there should appear the phone number
image

2) These two photos don't have the border-radius
image
image

3) Remove the autofill styles from the name and eamil inputs. Here a hint how you can you do it, https://css-tricks.com/almanac/selectors/a/autofill/
image

GUSILLUS (changes requested):
GJ! But:

1. Add some [alert logic](https://www.w3schools.com/tags/att_a_href.asp) to this button
image

2. Make the sidebar more relevant to your design. For example you don't have a login and store page replace them
image


3. Disable scroll in the menu
image

PR #1296, opened 2024-04-10
Layout B&O

dmytro-popovych:
[DEMO LINK](https://dmytro-popovych.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. disable textarea scaling
image

2. disable the scroll when the menu opens
image

3. you should fix that
image

4. add the favicon
image

etojeDenys (approved):
also better to add the cursor pointer to this button
image


but looks good for me

PR #1305, opened 2024-04-15
done

yaroslav5481:
https://yaroslav5481.github.io/layout_landing-page/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -61,27 +61,10 @@
             </a>
 
             <div class="top-bar__icons">
-              <div class="icon icon--phone">
-                <div class="top-bar__contacts">
-                  <a
-                    href="tel: +1 212-535-7710"
-                    class="top-bar__location"
-                  >
-                    The Met Fifth Avenue
-                    <br />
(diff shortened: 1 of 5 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Now it is not possible to hover over this block that appears, because it immediately disappears, this needs to be fixed so that it is possible to hover over this block
image

2. You need to change the autocomplete styles and remove default border when input is focused
image

3. Add hover effects for these clickable elements
image

4. Remove the phone numbers block on the mobile and tablet versions
image

volodymyr-soltys97 (changes requested):
GJ!
To improve:
1. Still not fixed from the previous review, you need to change the autocomplete styles
image

2. This block should appear only on hover over the phone icon itself
image

3. The images should not overlap each other when hovering
image

4. You need to make a call when user clicking here on the mobile and tablet versions
image

IvanFesenko (approved):
Well done

PR #1311, opened 2024-04-21
initial commit

PieCyberPie:
[DEMO LINK](https://piecyberpie.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. You need to fix the background color here
image

2. The images should not overlap each other when hovering
image

3. This section should not have indents on the right and left, but should fit the width of the browser window
image

4. You need to change the autocomplete styles
image

5. Add transition for hover effects here
image

maxim2310 (changes requested):
maybe you have forgot to deploy last changes, because all last mistakes still there

PieCyberPie:
ah, yes, mb, deployed it now, maybe wait a minute until it reload

maxim2310 (changes requested):

need to use svg for icons
image

PieCyberPie:
in module was said to pay attention to image size, because sometimes .svg can be more heavy, than .png ,jpg etc.
but ok. I'll do it, wait a minute

PieCyberPie:
and one more question, do you read comments in here?

SanyaBratashchuk (approved):

PR #1317, opened 2024-04-23
layout_landing-page

Svitlana-Vitryk:
[DEMO LINK](https://Svitlana-Vitryk.github.io/layout_landing-page/)


anastasiia-tilikina (changes requested):
Well done! 
Just a few things we need to improve here


1. Remove horisontal scroll
image

2. Pls add transition for hover effects on these links
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Need to change the autocomplete styles for the inputs
image

2. Add cursor pointer for this button
image

DarkMistyRoom (approved):
Good job! I approve, but please fix these two points
- limit the container width by setting max-width instead of min-width
image

image

PR #1318, opened 2024-04-24
Develop

imOlya:
[DEMO LINK](https://imOlya.github.io/layout_landing-page/)


TarasHoliuk:
1. The footer shouldn't be always visible as background: 
image

2. Add favicon

3. Would be nice to add some hover effect to these links (in addition to cursor pointer):
image

4. Inputs shouldn't "jump" on hover (change outline styles instead of adding border):

https://github.com/mate-academy/layout_landing-page/assets/86923089/b0088cba-7cd9-49c2-b865-866b6ea6b6cb

5. Make all fields required

6. Add cursor pointer to all such buttons and to all images:
image

TarasHoliuk (changes requested):
Nicely done 👍 
Let's improve it to make even better 🙂

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. The images should not overlap each other when hovering
image

2. Need to change the autocomplete styles
image

3. Fix placeholder text as on the design
image

4. Need to fix hover effects on the mobile version
image

nazarmatsevych (approved):
Good work!

PR #1320, opened 2024-04-25
add solution MyBike

OlyaVidzi:
[DEMO LINK](https://OlyaVidzi.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Need to fix font styles for the phone number
image

2. You need to fix font styles everywhere, check it on the design
image

3. Need to remove scroll when this menu is open
image

4. Need to change the autocomplete styles
image

5. Need to centered the background image on the tablet version
image

PR #1330, opened 2024-04-29
add solution

santana1337:
- [DEMO LINK](https://santana1337.github.io/layout_landing-page/)


vadiimvooo (changes requested):
Good job. A few things to fix:
1. Remove scroll when burger menu is opened
image
2. Add validation for email field
image

PR #1333, opened 2024-05-01
Develop

OksanaK24:
[DEMO LINK](https://OksanaK24.github.io/layout_landing-page/)


maxim2310 (changes requested):

1. would be nice to add cursor pointer on all interactive elements
image
2. on my opinion need to add max width for header 
image
3. check font family here, mov they are different 
image
4. remove autofill styles 
image
5. would be nice to add animation on showing menu

OksanaK24:
Hi, I did all of that, deployed again and now the page is fully broken on git-hub side. How to fix that?

l4st1m0za (changes requested):
If you have any questions you need to ask them in FE chat. In the GitHub menthors are not able to see your comments at the time you write it. So please, describe your problem in FE chat

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Logos and icons must be aligned horizontally
image

2. Add hover effects for all clickable elements and add indents between the phone numbers
image
image

3. Add the indent between the button and footer, check it on the design
image

4. You need to hide the phone numbers block on the mobile and tablet versions
image

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. Add hover effects here
image

2. Need to add transition for hover effects
image

3. If you now click on this empty area, then a call occurs everywhere, also this block should be shown only when hovering over the icon itself, because it now appears even when hovering over this empty area, this needs to be fixed
image

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. Now it is not possible to move the cursor on this block that appeared, because it immediately disappears, it needs to be fixed
image

PR #1338, opened 2024-05-04
Develop

SrTrace:
- [DEMO LINK](https://SrTrace.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. The background image should span the entire width of the browser window
image
image

2. This block you need to show when hovering the phone icon, and for the call no need to open the new tab
image

3. Need to remove the scroll when this menu is open 
image

4. The user cannot be able to change the size of textarea field 
image

5. Need to change the autofill styles 
image

6. These elements must be links and add hover effects here 
image

SrTrace:
[LandingPage](https://srtrace.github.io/layout_landing-page/)

It seems like all good now)

l4st1m0za (changes requested):
1) Phone numbers here should be clickable
image

2) Use .svg for icons

SrTrace:
[Landing Page]( https://srtrace.github.io/layout_landing-page/)

updated icons to svg format, made phone menu clickable

l4st1m0za (approved):
I will accept, but please make so that phone numbers appear on the hover effect, but they stay so that I can click on them. And when I get off that area, they should dissapeear

PR #1339, opened 2024-05-06
add solution

HumenVAsya:
[DEMO LINK](https://HumenVAsya.github.io/layout_landing-page/)


dinaminahi (changes requested):
Good job 👍 
But we need to do some fixes

Firstly I can't find some blocks and images that are in design, let's add them

This one looks odd, let's make it more like in design
image

Let's fix fonts, because they are different through your page

Look through [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md) and make sure you did it

HumenVAsya:
не розумію чому лінка веде на стару версію,ghs pfgecre npm start все що я додав коректно працювало

dinaminahi (changes requested):
Demo is not working 😢 
Have you run `npm run deploy` after pushing changes?

HumenVAsya:
I don't know why i don't see styles

roman-mirzoian (changes requested):
Don't worry if something doesn't work out, well done. Try to rebuild gh-pages if something doesn't work, ask for help in the qna channel.

dinaminahi (changes requested):
Good job 👍  I'm happy that the demo is working
I understand that you were busy deploying it, let's now spend some time improving it

There are few checklist items we need to do:
6. All addresses should be links to some location on Google Maps. Open them in a new tab with target="_blank".
```
}
```


```
.menu {
}
.menu__top {
  top: 0;
  z-index: 1;
  background-color: #e5e5e5;
}
```

HumenVAsya:
не зрозумів що я маю пофіксити тут 
'The page shouldn't be reloaded on form submit, but the fields should be cleared. Use:
<form onsubmit="this.reset(); return false;">'

roman-mirzoian (changes requested):
Good work, just a little more to do.

First, these images must also have a hover:
(7. Increase all images on hover (make them links to #, if there is no better option))
https://gyazo.com/9642b886232e5fdaa72ecbfcd8124875

Input field and buttons do not look the same as on the mockup:

image

image

roman-mirzoian (approved):
Good job! 
Be more careful next time, if something doesn't work, ask for help in the qna channel.

PR #1343, opened 2024-05-08
add task solution

arthurgaleta:
[DEMO LINK](https://arthurgaleta.github.io/layout_landing-page/)


l4st1m0za (changes requested):
1) Check indents in the Message field
image

2) Remove autofill styles in the inputs
https://css-tricks.com/almanac/selectors/a/autofill/

arthurgaleta:
1. Fixed indents 
2. Used trick from https://css-tricks.com/almanac/selectors/a/autofill/ , don't know if it's works, because I can't reproduce the problem, so for me it's looks the same as before.

anastasiiavorobiova (changes requested):
Great work! Consider making hover smooth here 
Screenshot 2024-05-08 at 14 43 13

arthurgaleta:
Fixed.

anastasiiavorobiova (changes requested):
Great work! The hover on the button should be smooth.
Mouseleave should be smooth also 
Screenshot 2024-05-09 at 14 15 41
Screenshot 2024-05-09 at 14 15 47

arthurgaleta:
Fixed.

VitaliyBondarenko1982 (approved):
Looks good.
Found a couple moments, which you can fix during prepare project for portfolio:
- set transition for selector (not for :hover). it will make smooth change styles in both directions.
chrome-capture-2024-5-11 (1)
- something wrong happened with inputs after autocomplete.
chrome-capture-2024-5-11

PR #1347, opened 2024-05-10
Develop

Elvira-Krupa:
- Deploy and create a Pull Request with a [DEMO LINK](https://Elvira-Krupa.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. looks different
image
image


2.
image
image



My suggestions:
1. better to disable the textarea scaling
image

2. better to import all icons as SVG, it will improve at least the quality
image

image

vadiimvooo (changes requested):
Good job. there is a few changes required:
1.  Add transition on hover effect for the phone links
image
2. Image should not overlap neighbors.
image
3. Add transition for the button as well
image
4. All interactive elements (links, inputs, etc) in this block should be with transition property.
image

PR #1349, opened 2024-05-13
add solution

yaroslav1177:
This is my project about museum. 
I used SCSS, BEM, some kind of transformation 
[DEMO](https://yaroslav1177.github.io/layout_landing-page/)
[LAYOUT](https://www.figma.com/design/lSR1m42L9YwzQwzzxKwHpw/THE-MET?node-id=8590-29&t=7HFOYHMWUJmfIwrR-0)


dinaminahi (changes requested):
Yaroslav, you dit it great 🥇  I see only these thinks you need to fix:

The menu is not one full height of screen on bigger width
image
image

And you need to make  form fields  **required**
image
image

yaroslav1177:
at first i did not see about required, i thought that something wrong with types of fields, but i understood and fixed

roman-mirzoian (changes requested):
Good job! Let's fix some points.

1) It is necessary to slightly correct the size of the background image of the header (a white stripe appears at the size of ~ 1200px)
https://gyazo.com/ac31aea957310566b74551364d0db6d1

2) I would also slightly reduce the hover size effect of the pictures so that they do not interfere with the text.
https://gyazo.com/de64f32e2a1f75c3d8f39ee3dc9784f0

roman-mirzoian (approved):
Well done!

PR #1358, opened 2024-05-15
Develop

Stasandrushchak:
[DEMO](https://stasandrushchak.github.io/layout_landing-page/)
[REPO](https://github.com/Stasandrushchak/layout_landing-page/tree/develop)


roman-mirzoian (changes requested):
Good job. but there is something to fix.

1) It is better to add a small hover size effect to the pictures
2) When the menu is open, you need to block scrolling with the mouse wheel.
image
https://gyazo.com/8c833853bb0dde79340b4e63ce6dc638

3) At the bottom, the contact information must be clickable with each type of link.
image


4) Input fields do not work correctly (the form should simply clear when the button is pressed)
image


your version - https://gyazo.com/6617bd156502fdbcbfe683ad462d2e8c

correct example - https://gyazo.com/412b3a33e8b9eda8460b367a85af1b03

roman-mirzoian (changes requested):
Good job, but I didn't see any changes in the demo, maybe you accidentally forgot to update deployment.
https://gyazo.com/5c0120bbc8e17ce10f72901c77775f0b

roman-mirzoian (changes requested):
Good work, I see positive changes. But there is still work to be done.

Let's go through the points of the [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md):

1) Use the logo as a favicon
image

2) Make sure phone icons and phone numbers are real links to start a call:
These links must be clickable
https://gyazo.com/8778f4f3f9dd1ed632d1d41a3ee57528

3) Increase all images on hover. It is better to add a small hover for all pictures:
https://gyazo.com/6950f9555173e016c46f9ea3fd30f460

4) It is better to add a link to the appropriate section in the menu link
https://gyazo.com/cdd7b7e0fb7580c041fffaa495348b0a

Only email space works here:
https://gyazo.com/0744cbed3405f70885c5e2e8bb2c0003
image

image

Also, the footer image has an unnecessary indent from the bottom
image

roman-mirzoian (changes requested):
Good work but there is still work to be done:

1) Need to add such a hover to the phone: 
https://gyazo.com/bfbcbe731eeb5b25a8b86ef8afce56bb
image

2) I also noticed the problem that after a width of 525 pixels, the display of the elements of the window breaks
https://gyazo.com/4897f37b187bc24a9c73283b6a164d7e

3) There is also the problem that the hints are only in the email field. 
Need, when entered only the name, a prompt appears that it is better to enter e-mail and vice versa with other fields : https://gyazo.com/c7101c6183bac4a2feb6b3884a7165b4

roman-mirzoian (approved):
Good work, there is still a small bug, here's how to fix it:
(on a small screen up to 320px, as in the mockup, the display of the page and input fields is a little broken (note that it is better to set it only for a small screen))
https://gyazo.com/80d5d7443a56fcac72632da3b8e4fe01

PR #1359, opened 2024-05-15
add task solution

demkoveronika:
[DEMO LINK](https://demkoveronika.github.io/layout_landing-page/)


roman-mirzoian (changes requested):
Good work, but there is room for improvement.
1) it is better to add a small hover size effect to all pictures.
2) at the size of 740 - 575 pixels, the background of the header breaks.
https://gyazo.com/e0e2447bc318bf1923c197aa9fa87d46

roman-mirzoian (changes requested):
Good work, but the problem with the header and footer background images remained and
because of this, horizontal scrolling appears.

1260 - 1025 px
https://gyazo.com/80921c96a1344500f23a213633fab16e

roman-mirzoian (approved):
Well done!

PR #1361, opened 2024-05-15
add task solution

OSpyrydonov:
[DEMO LINK](https://OSpyrydonov.github.io/layout_landing-page/)


roman-mirzoian (changes requested):
Well done, everything works right.
However, it is better to add a small hover size effect for all pictures so the site will look more dynamic.

roman-mirzoian (approved):
Well done!

PR #1367, opened 2024-05-15
add task solution

V-Shut:
[DEMO LINK](https://v-shut.github.io/layout_landing-page/)


roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

1) Use the logo as a favicon

image

2) The menu should open to the full height of the page, regardless of size.
And page scrolling should also be blocked when the menu is open:
https://gyazo.com/574db1af7dc21f52786df3243dc21458

3) It is better to add a small hover size effect for pictures.
Also added a scroll to the top of the page on click:


image

4) Сontact information should be clickable and with an appropriate link type:

image
image

5) image
your version - https://gyazo.com/e8a8c3212a24ed2444bbbed499c54d36
correct example - https://gyazo.com/c4b8854a9f09ee0a0933497313bc2bbf

6) The background of the header and footer at a size of 100 pixels or more breaks their display.

try to fix it so that everything looks good

image
image

dinaminahi (changes requested):
Good job Vitalii ⭐ 
Please, go through checklist and fix it in your code. I underlined all undone


image
image


I also can see these problems with horizontal scroll and header not on the full height
image

dinaminahi (changes requested):
Good job 👍 I see that you fixed a lot 💪 
But still not everything is done 😢 

image

Here are the videos how the form and footer should behave

https://github.com/mate-academy/layout_landing-page/assets/46566190/10ce75e2-0044-4efd-9c5e-a241a844ad6c


https://github.com/mate-academy/layout_landing-page/assets/46566190/acf11bb6-a4fd-4134-ae3c-fbd46f0d6293

roman-mirzoian (changes requested):
Good job, but there are a few remarks:

First, the background image of the header does not look right on a small screen.
image

Second, I suggest slightly reducing the hover size of the pictures so that they do not overlap each other:
https://gyazo.com/d413acdcccc6c9f92c77c844d0af851f

Also the link in the menu to the corresponding section does not work:
https://gyazo.com/6e40a9b2cf9fa0491afff693708623fd

V-Shut:
The menu was originally made from checkbox, so in order not to rework it, I had to find a script to change checkbox status. I apologize in advance)

roman-mirzoian (approved):
Well done!

PR #1375, opened 2024-05-19
add task solution

pasha28091997:
[DEMO LINK](https://pasha28091997.github.io/layout_landing-page/)


vadiimvooo (changes requested):
1. Remove scroll in menu
image
2. Phone hover effect is broken
image
3. Contact us section have different spacing. Please fix it according to mockup. Also on hover the elements in this section jump, please fix it as well
image
4. Height of this element is different, please fix it.
image
5. Fix it on mobile as well
image

vadiimvooo (changes requested):
Seems like not all comments were fixed:
1. The fields have strange paddings, please fix it
image
2. Images still not fixed
image
3. Please add hover effect for images as well
4. Add transition for the button
image

pasha28091997:
Как сделать 1. что бы надпись Message была сверху я не знаю( Подскажи пожалуйста 
@vadiimvooo

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page:
image

2. Remove this tag be and margin top for the first input 
image

3. The user cannot be able to change size of the textarea field 
image

4. You need to change the autofill styles for the form fields 
image

5. Need to reset form fields after sending form, also add a cursor pointer for this button 
image

image

7. The left indent must be are same, fix it. Also, you need to open this additional link in the new tab
image

8. Add hover effects for all clickable elements on the page 
image

9. You need to hide the phone number on the mobile and desktop versions 
image

10. Fix the left and right indents here, check it on the design 
image

11. The left and right padding must be 20px here 
image

vadiimvooo (approved):
Well done

PR #1376, opened 2024-05-19
Develop

jooaasss:
- [DEMO LINK](https://jooaasss.github.io/layout_landing-page/)


jooaasss:
I just noticed title problem and comment for you after my code, so I resolve all the issues after your comments and suggetstions

vadiimvooo (changes requested):
1. Remove scroll when menu opened
image
2. You don't need to refresh page on form submission

jooaasss:
> 1. Remove scroll when menu opened

  - can i do this by using js?

jooaasss:
while writing this code I also wanted to use nth-child selector in display grid, but it always applied for every grid, not the specific element.  how can i change styles for one element in my grid display? (for example row-gap) or is it even possible to change?

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. You need to make all these fields are required before sending 
image

2. Need to reset form fields after submit 
image

3. You need to hide the phone number on the mobile and tablet versions 
image

4. Need to center the background image 
image

volodymyr-soltys97 (changes requested):
.

volodymyr-soltys97 (approved):
Great work! 🔥

PR #1386, opened 2024-05-26
Add THE MET landing page

GoVolodya:
- Deploy and create a Pull Request with a [DEMO LINK](https://GoVolodya.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add hover effects for all clickable elements on the page 
image

2. Add cursor pointer for this button
image

3. Add transition for hover effects everywhere

4. You need to hide the additional phone numbers block on the mobile and tablet versions 
image

GoVolodya:
Done. I have made some changes. Pleas check this when you have time.

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. Where is the phone icon on the mobile and tablet? You need to hide only this additional block, not icon 
image
image

GoVolodya:
Well, that is make no sense for me to show icon if it's not doing anything, so I decided to hide it also. If that is bothering you I can easily fix that.

maxim2310 (changes requested):

1. check what is the line 
image
2. remove autofill style
image
3. these should be 2 links not 5
image
4.  when i click on phone while mobile nothing happens 
image

GoVolodya:
I have made some more fixes.

1st the red line is a part of menu which slides from left to right onclick. Somehow you caught it on screenshot. Please  provide steps to reproduce as for me its working fine (it uses 100vw for header and -100%  absolute positions so technically there should be no problems)

2nd I have added autocomplete attribute with value of off for the whole form which should prevent such behaviour theoretically. On practice some browsers may still have autocomplete as their feature and some don't have it at all even without attribute. (me personally think its a good feature which should be enabled)

3rd have combined links together so we have 2 of them instead of 5

4th The other mentor asked for such behaviour so I am not sure anymore who is correct at this point (please read PR comments for more details)

GoVolodya:
Have made some new fixes. It was good to know about autofill, never used that before, thanks =)

volodymyr-soltys97 (approved):
Great work! 🔥

PR #1388, opened 2024-05-26
task solution

mmivchenko:
[DEMO LINK](https://mmivchenko.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add a favicon and correct title on the page 
image

2. You need to show additional block when hovering this icon, check it on the design 
image

image

4. Add hover effects for images

5. The address must be links, also you need to add hover effects for all clickable elements 
image

6. These placeholder texts must be as placeholder, not as value 
image

7. Also, you need to add the placeholder text to the textarea and cannot be able to change size of the textarea field 

image

9. This part of the task is incomplete, you need to fix it
image

volodymyr-soltys97 (changes requested):
GJ!
To improve:
image

2. Now it is not possible to place the cursor on this block, because it immediately disappears, also when clicking on the icon, a call must be made, this must be fixed
image

3. The images shouldn't overlap each other when hovering, also you need to add transition for hover effects
image

4. You need to fix the inputs width, check it on the design 
image

5. The user cannot be able to change the textarea field 
image

6. You need to change the autofill styles and remove default blue border when input is focused 
image

7. You need to hide this additional block with phone numbers on the mobile and tablet versions 
image

VitaliyBondarenko1982 (approved):
Well done

PR #1389, opened 2024-05-27
Landing-page MyBiKE

AlShramchemko:
[DEMO LINK](https://AlShramchemko.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. You need to show the phone number when hovering this icon, check it on the design 
image

2. Don't need to use position for footer image 
image

image

4. You need to fix this error after click send in the form 
image

image

6. These elements must be as links and add cursor pointer here 
image

anastasiiavorobiova (changes requested):
Great work! 
Screenshot 2024-05-27 at 14 45 28
Consider removing default borders. Check the font here. A user shouldn't be able to resize the text area 
Screenshot 2024-05-27 at 14 39 55
Consider decreasing the size of the image on the hover. Also, consider exporting an image without the label

anastasiiavorobiova (changes requested):
Great work! Images overlap the text. Consider fixing it. 
Screenshot 2024-05-27 at 15 48 54
Form fields have different font families 
Screenshot 2024-05-27 at 15 48 35

anastasiiavorobiova (approved):
Great work!

PR #1395, opened 2024-05-28
Develop

Polina-7:
1. Use the logo as a favicon
3. All logos should be links to the top of the page
5. Make sure phone icons and phone numbers are real links to start a call
6. All addresses should be links to some location on Google Maps. Open them in a new tab with `target="_blank"`.
8. Make sure everything looks neat on mobile and without horizontal scrolling
9. The speed of animations should be the same throughout the page (for example, increasing when hovering or moving blocks when scrolling)
10. Ensure that all `nav__links` work smoothly
    ```css
    }
    ```
    ```css
    .menu {
    }
    .menu__top {
      top: 0;
      z-index: 1;
      background-color: #e5e5e5;
    }
    ```
13. All form fields should be required and have correct types (`email`, `tel`, etc.)
    ```html
    ```
16. Make the `footer` sticky at the bottom with `z-index: -1` to get a `fixed` background effect.

[DEMO LINK](https://Polina-7.github.io/layout_landing-page/)

### DIFF
--- a/src/styles/blocks/menu.scss
+++ b/src/styles/blocks/menu.scss
@@ -41,5 +41,8 @@
   }
 }
 
+.page:has(.page__menu:target) {
+  overflow: hidden;
+}
 
 
--- a/src/styles/blocks/page.scss
+++ b/src/styles/blocks/page.scss
@@ -29,8 +29,8 @@
           }
     }
 
-.page:has(.page__menu:target) {
-  overflow: hidden;
-  }
+// .page:has(.page__menu:target) {
+//   overflow: hidden;
+//   }
 }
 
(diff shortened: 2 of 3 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
image

image

3. The placeholder text must be on english
image

4. You need to change the autofill styles 
image

5. This element must be textarea 
image

6. You need to reset form fields after sending form 
image

SanyaBratashchuk (changes requested):
1. get rid of horizontal scroll

VitaliyBondarenko1982 (approved):
11 point from checklist not implemented.
Screenshot 2024-06-02 at 19 57 37

Suggest to fix it during prepare project for portfolio.

PR #1441, opened 2024-06-26
Solution

Galers:
Have a nice day!😊😊😊
[DEMO LINK](https://Galers.github.io/layout_landing-page/)


GUSILLUS (changes requested):
GJ!

1. Consider adding fixed height for header block similar to design  to avoid this situation
image

2. Consider fixing this logic
image

3. Consider disabled scroll when menu is open
image

4. Consider improve menu logic on a tablet and mobile
image
image

anastasiia-tilikina (commented):
GJ!
let's make a couple more fixes before final approval)

1. When I open menu and then press on logo or any link menu is closed but scroll is still disabled.
2. Add cursor pointer to this button
image
3. All external links (in footer) should be opened in new tab
4. position of the tooltip still needs to be fixed 
figma
image
your page
image
5. add scroll behaviour smooth for your page
6. I think spacing between these elements is smaller than in the design
figma
image
your page
image

anastasiia-tilikina (changes requested):
.

VitaliyBondarenko1982 (approved):
Looks good

PR #1452, opened 2024-06-30
Develop

Teollan:
[DEMO LINK](https://Teollan.github.io/layout_landing-page/)

This landing is based on [this](https://www.figma.com/file/lSR1m42L9YwzQwzzxKwHpw/THE-MET) figma design.

 ❗ **Rationale behind using inline styles in some `img` tags** ❗ 
Some images need unique `object-position`. Unique, as in 'cannot be shoved into a css class or reasonably changed with a modifier'. Obviously, same behavior can be achieved with id selectors, but as images are not standalone blocks, do not repeat, and do not have any children, therefore just writing an inline style (if necessary) seems better then writing overly specific `id` styles.


mrkovaliv:
Make mobile form bigger
Screenshot 2024-07-01 at 17 22 33

mrkovaliv:
Fix this issue:
Screenshot 2024-07-01 at 17 23 30

mrkovaliv (changes requested):
Good job 👍 👍 👍 

Added some comments))

Teollan:
Hi, thanks for the review!

I`ve definitely fixed the form stretch issue, but the second one is tricky. I could not replicate the issue consistently, but raising threshold for transition into desktop mode seems to help ¯\_(ツ)_/¯

BudnikOleksii (approved):
Good work 👍 
Approved, just one suggestion. All interactive elements such as buttons should have a cursor pointer

PR #1456, opened 2024-07-01
Landing page

victoriasokolvak:
[DEMO LINK](https://victoriasokolvak.github.io/layout_landing-page/)


BudnikOleksii (changes requested):
Good start 👍 
When done check yourself using the [CHECKLIST](https://github.com/victoriasokolvak/layout_landing-page/blob/develop/checklist.md) when finished;
- Use the logo as a favicon
- Show the phone number on hover phone icon:

image

image

image

- When a user clicks on "send" button or press "enter" in the form it should submit the form. But now the send button is actually a div with a small link inside

image

- Email should be a link with mailto so it opens some mailer. https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_link_mailto
- Background in this section doesn't cover all screen like in the design:

image

BudnikOleksii (changes requested):
Please, don't forget to deploy your solution so a reviewer can test the UI
image

BudnikOleksii (changes requested):
Looks good 👍 
Just a few more moments that should be fixed:
- All transitions should be smooth and have the same transition duration(check images on hover)
- Button in form has some shadow, not as in the design
image

BudnikOleksii (approved):
👍

PR #1465, opened 2024-07-01
add task solution

onikishyna:


src/styles/contact-us.scss
```diff
@@ -0,0 +1,205 @@
...
+  &:hover {
+    color: $main-dark;
+  }
+
+  &:active {
+    color: $main-dark;
+  }
+}
+
+textarea {
```
ArturErmolenko, line 147: add styles to classes not tags, plz fix it everywhere

ArturErmolenko (changes requested):
Plz add Demo link, we can't check UI ;(

ArturErmolenko (changes requested):
Add Demo link please 😢

BudnikOleksii (changes requested):
Good start, but a lot of work has to be done.
Please, check the design and the [checklist](https://github.com/onikishyna/layout_landing-page/blob/develop/checklist.md):
- When I hover the header in any place, the phone number appears. They should appear only when a user hovers an icon.
- The menu icon moves when I hover it.
- Sections looks different from the design
image

- Fix the horizontal scroll as well
image

- Inputs and textarea don't have paddings
image

- All interactive elements such as buttons/links/etc should have a cursor pointer

PR #1466, opened 2024-07-01
create landing page

0YehorVolkov0:
[DEMO LINK](https://0YehorVolkov0.github.io/layout_landing-page/)


ArturErmolenko (changes requested):
Great job 🔥 
Let's make some improvements:
1. When I click on phone it should make a call(now when I click nothing happens)
Screenshot 2024-07-02 at 12 55 11

2. When clicked on a section in the menu, the corresponding section should open (ideally so that there is a smooth scroll) now it just close menu
Screenshot 2024-07-02 at 12 55 41

3. Let's add transition on  hover  effects of pictures and buttons (now the pictures twitch when I remove the mouse) and the color of the button changes instantly.
All animations on the landing page should occur smoothly. Please check all places with animation.

4. This section should be interactive, that is, when clicked, a call should occur, mail or cards should open + let's also add hover effects on it
Screenshot 2024-07-02 at 12 56 19

PR #1470, opened 2024-07-02
add-landing_page-task-solution

vadimurbanovich:
[DEMO LINK](https://vadimurbanovich.github.io/layout_product-cards/)


BudnikOleksii (changes requested):
Looks good, just a few moments should be fixed:
- When a user hovers the phone icon you should show the phone number:
image

- All logos should be links to the top of the page(in your case when I click on the logo it calls, and the icon instead points to the top of the page, it should be vise verse)

BudnikOleksii (approved):
Good job 👍

PR #1480, opened 2024-07-06
add solution

Valentyn-Radobenko:
- [DEMO LINK](https://Valentyn-Radobenko.github.io/layout_landing-page/)

The names of the sections in the navigation menu differ from the design to match the sections on the page


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. You need to show phone number when hovering this icon 
image

2. Change the autofill styles for the form fields
image

3. You need to reset form fields after submitting form, also make all these fields are required before sending 
image

VitaliyBondarenko1982 (approved):
Looks good.
A couple moments which you can improve during prepare project for portfolio.
- increse each image separatlely
Screenshot 2024-07-08 at 08 08 25

- add transition for elements with hover effect to make changing styles smoothly
- inputs styles for autocomplete should be the same as without autocomplete.
Screenshot 2024-07-08 at 08 08 46

Valentyn-Radobenko:
> * increse each image separatlely

I originally incresed each image individually, but then I changed to incresing by category, because there should still be a link to one page, and it seems a little strange to me that you can select one of two images, but each of them leads to the same page. I understand that I'm not an expert in styleshitting, but I'm wondering why incresing of image should be separate, I'd like an explanation of this in more detail if possible

PR #1482, opened 2024-07-09
add task solution

Ilona-Gogol:
[DEMO LINK](https://Ilona-Gogol.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. You need to remove scroll when the burger menu is open 
image

2. When hovering here, all content jumps in this section, this needs to be fixed
image

3. The font styles for these placeholder text must be are same everywhere 
image

4. Fix the height for the textarea, check it on the mockup 
image

5. After submitting form you don't need to refresh page, you need only reset form fields 

6. Open this additional link in the new tab 
image

7. Hide the phone number on the mobile and tablet versions 
image

anastasiiavorobiova (approved):
Great work! A user should be able to scroll the menu in the landscape mode 
Screenshot 2024-07-11 at 15 38 30

PR #1483, opened 2024-07-09
Develop

2bmakohon:
[DEMO LINK](https://2bmakohon.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
image

2. Add hover effects for all clickable elements on the page

3. You need to show phone number when hovering this icon 
image

4. Add transition when this menu appears on the page 
image

5. The left indents must be are same here 
image

6. The left and right indents must be are same everywhere 
image

7. Add hover effects for images 
image

8. The user cannot be able to change the textarea field 
image

9. Make all these fields are required before sending 
image

10. These elements must be links 
image

PR #1485, opened 2024-07-10
Develop

vladisssi:
- [MYBIKE landing](https://www.figma.com/file/NZQAIydtHo5QkINyGLHNcq/BIKE-New-Version?node-id=0%3A1)
- [DEMO LINK](https://vladisssi.github.io/layout_landing-page/)


anastasiia-tilikina (changes requested):
Good job!
Let's make things better)

1. Add tooltip on phone icon hover
image

2. There is no upperCase for this title in the design
image

image


4. There is some gap between texts here pls add it to your page
image

5. Fix font-family for textarea and disable resize
image

6. Form shouldn't reload page on submit. It should only reset fields

7. Add some transition on menu appearance
8 Fix background position of hero image on mobile (check design)

volodymyr-soltys97 (changes requested):
Hi, the comments from the previous review still not fixed, you need to make a deploy after your changes and check it after on the demo link 

2.  The maximum width of the content should be larger than 1024px, because it does not look good on large screens
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add transition when the burger menu appears on the page
image

image

3. Need to remove scroll when the burger menu is open 
image

4. Fix the border radius for the inputs and button width, check it on the layout 
image

5. Change the autofill styles for the form fields 
image

image

7. Add cursor pointer here 
image

8. Add hover effects here 
image

9. This address must be link 
image

10. Hide the phone number on the mobile and tablet versions 
image

11. Change the background image position to 90% 
image

12. The menu should scroll when the content does not fit
image

PR #1491, opened 2024-07-14
add task solution

vitorhvieira:
- Deploy and create a Pull Request with a [DEMO LINK](https://vitorhvieira.github.io/layout_landing-page/)


src/index.html
```diff
@@ -1,22 +1,390 @@
...
+                  class="contacts__input contacts__input--message"
+                  name="message"
+                  placeholder="Message"
+                  required
+                ></textarea>
+                <button class="button contacts__button">Send</button>
+              </form>
+
+              <div class="contacts__info">
+                <div class="contacts__info--block">
```
pedro-ruas, line 349: I believe you wanted an *Element* called *info-block* (with only 1 hyphen). Two hyphens means it's a modifier.
In this case, I don't see `block`, `title` and `text` as being modifiers, but part of the element name

src/styles/blocks/about.scss
```diff
@@ -0,0 +1,54 @@
...
+
+  &__title {
+    margin: 0;
+    grid-column: 1 / -1;
+
+    @include on-tablet {
+      grid-column: 1 / 4;
+    }
+
+    @include on-desktop {
```
pedro-ruas, line 32: Nice use of mixins!

pedro-ruas (approved):
Great code! Very well done ;D

PR #1516, opened 2024-07-27
Develop

2bmakohon:
[DEMO LINK](https://2bmakohon.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Fix the background image as on the layout 
image

image

image

4. This element must be link to mail 
image

5. Add hover effects here
image

6. This element must be as form tag, not div 
image

SanyaBratashchuk (approved):
1. apply transitions for hover effects

PR #1528, opened 2024-08-03
layout_landing-page

blacktox96:
[DEMO LINK](https://blacktox96.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good start!
To improve:
1. Add the left and right padding for the content 
image

2. Need to show additional block when hovering this icon 
image

3. Add transition when the burger menu appears on the page.
Also, you need to remove scroll when this menu is open 
image
image

image
image
image

blacktox96:
> 1. Add the left and right padding for the content
- My inline paddings set as they should be (like on Figma markup)

> 2. Need to show additional block when hovering this icon
- Nothing about this requirement in task or checklist

>3. 
      1. Add transition when the burger menu appears on the page.
      - Nothing about this requirement in task or checklist
      - 
      2. Also, you need to remove scroll when this menu is open
      - Done ✅

>4. Add hover effects for all clickable elements
- Done ✅

TarasHoliuk:
> Need to show additional block when hovering this icon
Nothing about this requirement in the task or checklist 🧐

1. Yep - it isn't mentioned in the checklist, but there are a lot of elements in the landing which is not described in detail in the checklist. But it's part of the mockup, so it should be implemented:
image

Anyway in the future you will add this landing page to your portfolio, so it's better for you to make it look more fancy 🙂 

2. Similar with this point:

>3.1. Add transition when the burger menu appears on the page
Nothing about this requirement in task or checklist 🧐

It's a general rule and will be mentioned in the next videos (if wasn't yet). All changes (like animations, changes on hover, focus, and actions like menu appearing) should be smooth
But this topic is one you will learn in the future. So when you will complete work on the landing page take these points ⬆️ into account.

For now, it looks good. Anyway, I reject it (that's a special flow of this task - it'll be approved when finished in future task). Now just move on in the course to the next topic

TarasHoliuk (changes requested):
GJ 👍

blacktox96:
@TarasHoliuk
Thanks for explanations

PR #1531, opened 2024-08-03
add task solution

VovaMandzii:
https://VovaMandzii.github.io/layout_landing-page/


GUSILLUS (changes requested):
GJ! 

1. Consider removing paddings
image

2. Consider showing clickable numbers when you hover a phone icon similar to design
image

3. Consider adding max-width property to fix UI of elements
image

4. Maybe you forgot some margin and text styles
image

5. Same with text style
image

6. UI BUG
image

7. Hover bug
image

8. Make contact block similar for design
image

etojeDenys (changes requested):
image
2. change the title of the page and add some favicon
image
3. button smaller then fields
image

4. add some hovers to all links
image

5. 
image

6. number should appear on hover
image

7. looks different
image
image


8. also you need to fix your fonts, cause it's different

SanyaBratashchuk (changes requested):
1. input should have paddings 
image
2. apply `cursor: pointer` for all interactive elements
3. background should be full-width
image

etojeDenys (approved):
lgtm

PR #1545, opened 2024-08-11
add task solution

vovaAlieksieiev:
[DEMO LINK](https://vovaAlieksieiev.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. You need to show the phone icon here, check it on the layout 
image

2. Don't need to use position for the header 
image

3. Fix default color for this button 
image

4. Remove default blue border when input is focused 
image

5. Use SVG for icons 
image

PR #1548, opened 2024-08-14
deploy for mentors advice

DimaDamage91:
[DEMO LINK](https://DimaDamage91.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. don't forget to add a favicon
image


2. disable the scroll when menu is open
3. implement all sections and after that request the review
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. You need to show additional phone numbers when hovering this icon, check it on the layout 
image

2. Remove scroll when the burger menu is open 
image

3. The user cannot be able to change the textarea field 
image

image

5. Add cursor pointer and transition for hover effects here 
image

6. These elements must be links and add hover effects here 
image

7. After submitting form you need reset form fields and don't need scroll page to top 
image

8. Remove default blue border when input is focused 
image

GUSILLUS (changes requested):
GJ!

1. User need ability to click on this numbers
image

2. Consider adding hover effects
image

3. Same 
image

4. Same with links
image

volodymyr-soltys97 (changes requested):
Good job!
To improve:
image

image

3. After the hover effects, the image should decrease as smoothly as it increases when hovering

4. It should be one link, it should not be broken down into lines, and it should also lead to a current address on Google Maps
image
image

image

6. You need to make a call when user is clicking on this icon 
image

TarasHoliuk (approved):
All previous comments resolved - approved ✔️ 

In addition, before adding it to the portfolio (don't worry - there will be a separate task for it before employment) would be nice to add a few small improvements:

1. Do not add a border on focus input/textarea - just change its color from transparent to black. Because now content below "jumps" on focus.

2. Add transition for color change on hover for links in "Contact Us" section.

3. Adjust image position in the mobile version:
image

P.S. No need to re-request review, task is already approved. Just suggestions to improve it before adding to the portfolio 🙂

PR #1553, opened 2024-08-15
Develop

CamilleHaus:
DEMO LINK - https://CamilleHaus.github.io/layout_landing-page/


pedro-ruas (changes requested):
Camille, the styling is as planned, but there is some room for improvement.
- Not all transitions are the same or as smooth as others. Images resizing, for example, could be smoothed
- Email links are not working, they should work similar to the phone call, triggering an email sender
- You could have more points where the page scrolls so we can test the smooth-scroll effect (like clicking the "explore" button, could lead to the header or to the products and so on)
- When images get scaled-up on hover, they are overlapping each other or the text from other items, you could either add more spacing or use less scaling so it doesn't happen
- When hovering links inside the menu, the menu itself is jumping, because of the new border added. You can use a transparent border when not-hovering and change only the color on hover, to prevent this from happening

gitlherme (approved):
Really great job, @CamilleHaus! 🧡😊

PR #1558, opened 2024-08-16
add task solution

joselangoni:
- [DEMO LINK](https://joselangoni.github.io/layout_landing-page/)


pedro-ruas (approved):
Jose, your code is very well done, you made great usage of SCSS!
Only thing that I can suggest as improvement in the future, is that, for "class-like" @mixins, you can use %classes and @extend instead.
In this code, for example, `@mixin button` and `@mixin input` could be replaced by `%button` and `%input`.
Then, when using it, instead of `@include button` you would use `@extend button`.
This is a small difference, but the resulting CSS (after SCSS gets compiled to CSS) gets slightly different and more compact when using @extend.
You can check more about it here: [@extend vs @include](https://vanseodesign.com/css/sass-mixin-or-extend/)

joselangoni:
Hi @pedro-ruas , thank you so much!
I will check what you suggested and I will definitely use it this way in the future.

PR #1559, opened 2024-08-16
layout_landing-page

davicatarino:
[DEMO LINK](https://davicatarino.github.io/layout_landing-page/)


pedro-ruas (approved):
Great job, Davi! Nice usage of SCSS!
Only thing that I can suggest as improvement in the future, is that, for "class-like" @mixins, you can use %classes and @extend instead.
In this code, for example, `@mixin page-button` could be replaced by `%page-button`.

Then, when using it, instead of `@include page-button` you would use `@extend %page-button`.
This is a small difference, but the resulting CSS (after SCSS gets compiled to CSS) gets slightly different and more compact when using @extend.
You can check more about it here: [@extend vs @include](https://vanseodesign.com/css/sass-mixin-or-extend/)

PR #1567, opened 2024-08-20
finish page

ekateriina7:
https://ekateriina7.github.io/layout_landing-page/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. You need to show additional phone number when hovering this icon, but only on the desktop 
image

2. Need to remove scroll when the burger menu is open 
image

image

4. Fix border radius for the textarea 
image

5. Change the autofill styles for the form fields
image

6. When hovering, fields in the form should not jump a few pixels up and down

7. Fit the background image here, check it on the layout 
image
image

etojeDenys (approved):
looks good

PR #1572, opened 2024-08-22
landing page

EustasTheMonk:
[DEMO LINK] (https://EustasTheMonk.github.io/layout_landing-page/)

Added selector ' * ' expecting the same behavior from all elements


BudnikOleksii (changes requested):
Good work 👍 
- Background color in the menu is different from what we have in the design
image
- Link with email should be mailto so when you click it opens the corresponding app, https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_link_mailto
image
- The textarea has wrong paddings or border radius
image

VolodymyrKirichenko (changes requested):
not fixed(Bug with colors). If u already fixed it - deploy it again
Знімок екрана 2024-08-26 о 16 30 28

VolodymyrKirichenko (changes requested):
Overall - looks good, but review my comments and fix them 🥹

PR #1573, opened 2024-08-22
add Landing-page

viktoribu:
[DEMO LINK](https://viktoribu.github.io/layout_landing-page/)


BudnikOleksii (changes requested):
Good work 👍
Just a few suggestions:
- You didn't set the max size of containers where you place the content, so on big screens the content stretches
image
- Adjust the position of the header and burger menu
image
- Link with email should be mailto so when you click it opens the corresponding app, https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_link_mailto
- All interactive elements such as links/button should have cursor pointer

PR #1578, opened 2024-08-23
add solution

KrivorotkoYurii:
[DEMO LINK](https://KrivorotkoYurii.github.io/layout_landing-page/)

Please don't treat a block with a container class as a BEM - block, instead treat it as a normal DOM element with some restrictions. The container element is not part of the BEM, we use it to limit the width.


src/index.html
```diff
@@ -16,7 +35,400 @@
...
+              class="top-bar__logo-link"
+            >
+              <img
+                src="images/logo-mobile-white-2x.png"
+                alt="MyBike logo"
+                class="top-bar__logo"
+              />
+            </a>
+            <div class="top-bar__icons">
+              <a
```
VolodymyrKirichenko, line 55: i don't see this logic on `hover`
Знімок екрана 2024-08-26 о 15 26 44

VolodymyrKirichenko (approved):
Good job! 🔥

PR #1579, opened 2024-08-23
results

maxmodrr:


etojeDenys (changes requested):
1. you should show number on icon hover
image

2. color looks broken
image

3. add the cursor pointer to the button
image

etojeDenys (approved):
lgtm

PR #1603, opened 2024-08-28
B&O landing page

vkrasnovyd:
- [DEMO LINK](https://vkrasnovyd.github.io/layout_landing-page/)
- [PAGE PROTOTYPE](https://www.figma.com/file/DtkQmQ797hk0nI4KfMi2Uq/BOSE-New-Version?type=design&node-id=6817-212&t=ZTV6Gl8NzaWkJ4FK-0)


anastasiiavorobiova (changes requested):
Great work! A user shouldn't be able to resize a text area.
Also, consider showing the phone number on the hover. 
Screenshot 2024-08-28 at 17 01 33

etojeDenys (changes requested):
1. still not fixed, you should show the number on hover, like in the figma
image


2. we don't need underline here
3. "Google Maps" text also looks redundant here
image

etojeDenys (approved):
🤩

PR #1618, opened 2024-09-05
add task solution

Mykolyan:
[DEMO LINK](https://Mykolyan.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job! 
Let's improve your page
1. The price must be are on the same line 
image

2. The user cannot be able to change size of the textarea field 
image

3. Change the autofill styles for the form fields and remove default blue border when input is focused
image

4. Add transition for these hover effects 
image

image

GUSILLUS (changes requested):
GJ!

1. Consider adding max-width to improve ui on a big screen

Mykolyan:
> 1. Consider adding max-width to improve ui on a big screen
Used to use max-width for desktop 1020px, inside content-padding-inline mixin, like in lecture video, but i didn't liked, how it looked on my widescreen monitor, and set  max-width: 1920px. Should I set it back to 1020px then?

SanyaBratashchuk (approved):

PR #1622, opened 2024-09-05
MyBike Landing

Olena7202:
[DEMO LINK](https://Olena7202.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. The menu should stretch to the full height of the browser window and scroll on the page when the menu is open
image

2. Fix font styles for the phone number 
image

3. Need to make a call when user clicks on the phone icon 
image

image

5. The font styles must be are same everywhere and remove default blue border 
image

image

7. After submit form don't need scroll page to top

image

9. This element must be link and add hover effects here

PR #1636, opened 2024-09-09
Develop

KomodomFS:
https://komodomfs.github.io/layout_landing-page/#


pedro-ruas (changes requested):
Ivo, I believe you did not read the checklist carefully. The Checklist explicitly says to NOT implement this design, and instead peak one of the four other options.
Pick one of your choice and do the implementation, please.
Your code is well done, but this task has a lot more to it, that you'll only be able to do with another template

PR #1638, opened 2024-09-10
Develop

an-marryKyslenko:
[Demo](https://an-marrykyslenko.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
image

2. It should be possible to select a phone number
image

image

4. Fix this block as on the design 
image

Design 
image

5. All these elements must be links 
image

6. Add cursor pointer for this button 
image

7. Make all these fields are required before sending and after submit don't need scroll page to top 
image

8. Need to make a call when user clicks here 
image

anastasiiavorobiova (changes requested):
Great work! 
Screenshot 2024-09-11 at 15 05 46
Check the alignment on the mobile
Screenshot 2024-09-11 at 15 05 33
The page under the menu shouldn't be scrollable 
Screenshot 2024-09-11 at 15 04 48
Consider using a link to Google Maps here. Check the color and the underline here

PR #1639, opened 2024-09-11
Landing the Met

Victoriia-Melnyk:
[DEMO LINK](https://Victoriia-Melnyk.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
image

image

image

4. Fix gap between the images, check it on the layout 
image

5. Add cursor pointer for all buttons 
image

6. The user cannot be able to change the size of the textarea field 
image

7. Fix gap between the links and forms 
image

volodymyr-soltys97 (changes requested):
Hi, there are no styles on the demo link, you need to make a deploy again and check it after on the demo 
image

PR #1640, opened 2024-09-12
done

Victoriia-Melnyk:
[DEMO LINK](https://Victoriia-Melnyk.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. The user should be able to hover over these phone numbers
image

image

3. Add gap between the images 
image

4. Need to centered these images 
image

PR #1656, opened 2024-09-20
finished project

CNegruzzi:
A project of a Nothing(R) tech website.

The website was built using such technologies as html, scss and js, adding to it Animate.css library and wow.js for nicer animation effects.

[DEMO LINK](https://cnegruzzi.github.io/layout_landing-page/)

- [Nothing](https://www.figma.com/file/DtkQmQ797hk0nI4KfMi2Uq/BOSE-New-Version?type=design&node-id=6802-139&t=L7eKz5YKLN0m5WxR-0)

### DIFF
--- a/src/styles/blocks/form.scss
+++ b/src/styles/blocks/form.scss
@@ -25,6 +25,7 @@ select:-webkit-autofill:invalid {
   &__form {
     margin-bottom: 48px;
     grid-column: 1 / -1;
+    cursor: text;
 
     @include on-tablet {
       margin-bottom: 80px;


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Need to remove scroll when the burger menu is open 
image

2. The prices must be are on the same line, check it on the design 
image

3. Change the autofill styles for the form fields and remove default blue border when input is focused
image

image

5. After submitting form don't to refresh page and scroll to top, you need only reset form fields

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. Remove default blue border when input is focused 
image
image

2. This phone number should appear and disappear smoothly on the page
image

etojeDenys (approved):
looks awesome, but better to fix these small jumps

https://github.com/user-attachments/assets/f4852f7d-be2d-4800-9819-a0ff83bffe46

PR #1659, opened 2024-09-22
Develop

lthmslf:
https://lthmslf.github.io/layout_landing-page/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Remove the default blue border when input is focused 
image

2. Fix the background image position, the top cyclist must not be clipped, check it on the design 
image

3. Use `1fr` for size grid columns instead of 80px 
image

image

volodymyr-soltys97 (approved):
Well done! 🔥

PR #1662, opened 2024-09-23
Create header and aside

AlyonaV22:
- [DEMO LINK](https://AlyonaV22.github.io/layout_landing-page/)


BudnikOleksii (changes requested):
Good start, but a lot of things from the design are not implemented

BudnikOleksii (changes requested):
Good work 👍 
Some moments should be implemented in accordance with a [checklist](https://github.com/AlyonaV22/layout_landing-page/blob/develop/checklist.md):
- When user hover icons you can see a different behavior in the design:
image

- Change text color on hover for phone, email, and address(also they should be links)
image

image

BudnikOleksii (changes requested):
Almost done 👍 
Menu looks strange
image
image

BudnikOleksii (approved):
Looks great 🤩

PR #1664, opened 2024-09-24
add solution

KateBerd:
[DEMO LINK](https://KateBerd.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
image

2. You need to show additional phone number when hovering this icon, check it on the design 
image

3. Remove scroll when this menu is open 
image

4. Fix the font weight for the prices 
image

image

6. The user cannot be able to change the size of the textarea 
image

7. Add cursor pointer for this button 
image

8. Make all these fields are required before sending 
image

9. These elements must be links and add hover effects here 
image

anastasiiavorobiova (approved):
Great work!

PR #1671, opened 2024-09-27
Add solution

azazeka:
[Demo](https://azazeka.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add a favicon on the page 
image

2. You need to show additional phone number when hovering this icon 
image

image

image

5. Fix the default background color for this button 
image

image

7. This element must be textarea, not input 
image

8. Add hover effects for these elements 
image

9. Fix gap between the button and links on the tablet and mobile versions 
image

PR #1676, opened 2024-10-01
add task solution

LX-y-ka:
- Deploy and create a Pull Request with a [DEMO LINK](https://LX-y-ka.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add a favicon on the page and change title to `The Met` 
image

image

image

4. Add hover effects for these clickable elements 
image

image

6. Hide the additional phone numbers block on the mobile and tablet versions, also, you need to make a call when user clicks on the phone icon 
image

SanyaBratashchuk (approved):
1. make all fields required

PR #1681, opened 2024-10-04
Add solution landing nothing!

DevMichelleBarcelos:
# Nothing Landing page: https://DevMichelleBarcelos.github.io/layout_landing-page

This landing page for Nothing store. It is present stores recommended products, also allows to choose products by category.
This page also provides "about us" section, and a form, for contacts.
Overview Layout is adaptive and responsive, implemented with the mobile-first approach, works correctly on all screen sizes (mobile, tablet, desktop).

# Techonoligies used

- **HTML**: Markup language for creating web pages.
- **CSS**: Cascading Style Sheets for styling web pages.
- **SASS(SCSS)**: extension for using variables, nested rules, grid conditional operators, mixins etc. Managing animations and transitions.
- **BEM**: Methodology for organizing classes. This makes the code more readable and understandable, helps to avoid naming conflicts and makes the code easier to maintain and reuse.

# Landing preview

[DEMO LINK](https://DevMichelleBarcelos.github.io/layout_landing-page/)

# Figma design

- [Nothing](https://www.figma.com/file/DtkQmQ797hk0nI4KfMi2Uq/BOSE-New-Version?type=design&node-id=6802-139&t=L7eKz5YKLN0m5WxR-0)


src/styles/blocks/about.scss
```diff
@@ -0,0 +1,34 @@
...
+      grid-column: span 6;
+    }
+  }
+
+  &__content {
+    grid-column: 1 / -1;
+    font-family: $secondary-font-family;
+    font-weight: 400px;
+    font-size: 16px;
+    line-height: 22.4px;
```
jv-aquino, line 24: avoid using half pixels (they can generally cause problems and sometimes are just not counted, i can only think of them working in box-shadow)

jv-aquino, line 24: avoid using half pixels (they can generally cause problems and sometimes are just not counted, i can only think of them working in box-shadow)

jv-aquino (approved):
wow michelle.... just... woww!!!!!! What a nice job! This really looks like a pro front-end

there are 2 tips that I would give you to improve your notions even more:
1 - avoid using half pixels - round them (generally to the nearest even number)
2 - try to use preloads for a better ux - in the 1st time, the animations were not behaving the way they should
try to search a bit more about these 2 topics and I guarantee it'll be worth!

PR #1684, opened 2024-10-06
add project

clavigo:
[DEMO LINK](https://clavigo.github.io/layout_landing-page/)


anastasiiavorobiova (changes requested):
Great work! 
Screenshot 2024-10-06 at 16 36 03
Consider removing default outlines
Screenshot 2024-10-06 at 16 35 51
Consider adding hover styles here 
Screenshot 2024-10-06 at 16 35 25
Consider showing the phone number on the hover

volodymyr-soltys97 (changes requested):
GJ!
To improve:
image

2. The prices must be are on the same line 
image

3. Fix default background color for this button 
image

image

TarasHoliuk (approved):
GJ! All previous issues were fixed - Approved ✔️ 

One small comment (you can fix it before adding landing to the portfolio, later):
All clickable elements should have `cursor: pointer` (buttons like View All, Send, etc.)

PR #1700, opened 2024-10-13
add task solution

lyarosh:
[DEMO LINK](https://lyarosh.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
image

2. You need to show additional phone number when hovering this icon and make a call when user clicking here 
image

3. Fix the width for the menu and remove scroll when menu is open 
image

4. Fix font styles here 
image

5. You need to check fonts everywhere 
image

6. The prices must be are on the same line 
image

7. Add hover effects for images 
image

8. Add the left padding for the placeholder text 
image

9. Change the autofill styles for the form fields
image

10. The user cannot be able to change size of the textarea field 
image

11. Fix default background color for this button 
image

12. These elements must be links and add hover effects here 
image

anastasiiavorobiova (changes requested):
Great work!
- Screenshot 2024-10-24 at 18 02 00
- Check the distance between sections for different devices
- Screenshot 2024-10-24 at 18 01 51
- The image shouldn't overlap the text on the hover
- Screenshot 2024-10-24 at 18 01 25
- Consider removing default outlines and borders
- The click on the submit button should clear the form

etojeDenys (changes requested):
1. looks different
image
image

2. remove it
image


3. hover effect shouldn't have these small jumps

https://github.com/user-attachments/assets/9bfbeae9-dcea-4467-af28-8c01427781b1


4. button should have the same width as other fields
image


5. font family looks different
image
image

etojeDenys (changes requested):
cool, now move this button inside the form tag
image

but don't forget to check, that form should not reload a page

VitaliyBondarenko1982 (approved):
Looks good

PR #1702, opened 2024-10-14
Develop

Artemida1609:
[DEMO LINK](https://Artemida1609.github.io/layout_landing-page/)


anastasiiavorobiova (changes requested):
Great work!
Screenshot 2024-10-14 at 15 44 32
Consider disabling the scroll under the menu
Screenshot 2024-10-14 at 15 43 45
- A user shouldn't be able to resize the textarea 
- All fields should be required
- The form should be cleared on the submit
Screenshot 2024-10-14 at 15 43 35
Consider using links here
Screenshot 2024-10-14 at 15 43 04
The phone number should be shown on one line

volodymyr-soltys97 (changes requested):
Good job! 
To improve:
1. You need to make a call when user is clicking on the phone icon 
image

image

3. The prices must be are on the same line 
image

4. Add a some hover effects for images 
image

5. Add cursor pointer for this button and add transition for hover effects here 
image

6. All these elements must be are links and add some hover effects here 
image

anastasiiavorobiova (changes requested):
Great work!
Screenshot 2024-10-15 at 15 26 31
Consider fixing previous comments
Also, the page is not scrollable after clicking on a link in the menu

Artemida1609:
Hello! I don`t have such an issue with phone number, I have in one line when i check it

anastasiiavorobiova (approved):
Great work!

PR #1717, opened 2024-10-17
Implement landing page

GrishaSeniv:
[DEMO LINK](https://GrishaSeniv.github.io/layout_landing-page/)


BudnikOleksii (changes requested):
Good work 👍 
- When a user hovers over the phone icon, the phone number should appear:
image

The form doesn't work actually, when I submit nothing happens
image

- And it has a slightly different design
image

PR #1721, opened 2024-10-18
Solution

ab3MN:
#### Implemented landing-page  `The MET`

-  [CHECKLIST](./checklist.md)
-  [DEPLOY](https://ab3MN.github.io/layout_landing-page/)
- [MET landing](https://www.figma.com/file/lSR1m42L9YwzQwzzxKwHpw/THE-MET)


src/styles/abstract/_mixins.scss
```diff
@@ -0,0 +1,64 @@
...
+  content: '';
+  display: block;
+  position: absolute;
+  height: $height;
+  background-color: $bgc;
+  margin-top: auto;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  animation: line-view 1s ease-in-out;
```
VolodymyrKirichenko, line 24: 1s is too long...
I think - (0.3-0.5) will be good

src/styles/base/_base.scss
```diff
@@ -0,0 +1,55 @@
...
+
+img {
+  display: block;
+  width: 100%;
+  height: 100%;
+  cursor: pointer;
+  transition: 1s ease-in-out;
+
+  &:hover {
+    transform: scale(1.2);
```
VolodymyrKirichenko, line 43: If you want to keep the scale on the image itself, not on the container, make the scale smaller. Right now, the picture cropped, and we can't see all the details when it's zoomed in

but it is better to change ur logic. After all, we do not see the whole picture

VolodymyrKirichenko (changes requested):
Well done 🔥 
but let's make some small corrections

src/styles/base/_base.scss
ab3MN: Hi, does this task require: "Increase all images on hover (make them links to #, if there is no better option)", or should I remove the overflow hidden? I thought it's a bad practice when the image will overlap other blocks

VolodymyrKirichenko: > Hi, does this task require: "Increase all images on hover (make them links to #, if there is no better option)".I thought it's a bad practice when the image will overlap other blocks

yes, u're right)
so u can make the scale smaller and everything will be beautiful)
**scale: 1.1**

Знімок екрана 2024-10-18 о 17 23 04

VolodymyrKirichenko (approved):
Well done 🔥 
Don't forget to check my comment 🥹

PR #1734, opened 2024-10-23
bose layout complete

ZupaFly:
[DEMO LINK](https://zupafly.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. disable the scroll if the menu is open
2. button shouldn't be stretched on desktop
image

3. let's follow the rules in next tasks
image

ZupaFly:
> 1. disable the scroll if the menu is open
> 2. button shouldn't be stretched on desktop
> 
> image
> 3. let's follow the rules in next tasks
> 
> image



> 1. disable the scroll if the menu is open
> 2. button shouldn't be stretched on desktop
> 
> image
> 3. let's follow the rules in next tasks
> 
> image

Oh, i didn't know that it was design from video. I just joined this study group from international group. And for us it was one of the designs from making your own layout. Because our studying video was based on Miami layout design. Ok, thats a new studying group and new rules, i will choose new design and will make new layout.

vadiimvooo (changes requested):
Good job, a few things to fix required:
1. Hover of element should not overlap the neighbour
image
2. Remove autocomplete default styles from inputs
image

TarasHoliuk (approved):
You have fixed the previous comments - approved ✔️ 

P.S. The border radius for the textarea should be 16px and not 50%. You can fix on your own - no need to re-request review- the task is already approved 🙂

PR #1735, opened 2024-10-23
add solution Landing Page -Bose

PedroWeb10:
[DEMO LINK](https://PedroWeb10.github.io/layout_landing-page/)


pedro-ruas (changes requested):
Pedro, I believe you have misundertood the task. Take a closer look in the guide and you will see that it says **explicitly** _not to implement the Bose page_ and **pick one of the four different templates**.
Besides that key point, I also must point out that there are other steps that must be followed, like: 
- Make sure all links work and send to the right part of the page,
- All images have a small animation when being hovered, increasing in size,
- The contact form should only be sent when every field is valid,
- Telephone and Address links sent to external sources accordingly,
- All page animations should have the same duration and be smooth,

PedroWeb10:
[DEMO LINK](https://PedroWeb10.github.io/layout_landing-page/)

pedro-ruas (changes requested):
Did you read my last comment? Why is your work still displaying Bose's page?

PR #1738, opened 2024-10-23
Landing Page project

AnderZerfall:
# Landing page

> Made for Mate Academy Task. Topic **Landing**.

The main goal of this project was to practice new skills I learned during Mate Academy courses.
To be more specific, I was focusing on understanding BEM technology and mastering SCSS preprocessor and responsive design.

There still might be some issues with the layout, but it works fine and smooth.

# See demo:

Check out the result here:

- [DEMO LINK](https://anderzerfall.github.io/layout_landing-page/src/index.html)

# Link to the design:

I took the following design as a reference:


BudnikOleksii (changes requested):
Please, deploy your solution so a reviewer can check the UI
image

AnderZerfall:
I'm so sorry about that. I have deployed the page, but GitHub removed it after a few days.

Anyway, I redeployed it. Could you check the link in the comment below? It's supposed to work fine now.

[DEMO]https://anderzerfall.github.io/layout_landing-page/src/index.html

BudnikOleksii (changes requested):
Looks good 👍 
Just a few suggestions and moments from the checklist:
- container should have max-width, as on the big screens 80% don't work well
image

image

- Break point for desktop is better to do at 1024, as a lot of tablets/mini laptops have resolution 768-1024. Also, you can check common resolutions for example here https://www.browserstack.com/guide/responsive-design-breakpoints
image

BudnikOleksii (approved):
Good work 👍 
Approved, but why is the menu centered?
image

PR #1739, opened 2024-10-24
add task solution

koguts:
міксини записала по іншому бо сварилось і не проходило 
1:1  ✖  Expected mixin name to be kebab-case    scss/at-mixin-pattern
   7:1  ✖  Expected mixin name to be kebab-case    scss/at-mixin-pattern
  13:1  ✖  Expected mixin name to be kebab-case    scss/at-mixin-pattern
  38:1  ✖  Expected mixin name to be kebab-case    scss/at-mixin-pattern


BudnikOleksii (changes requested):
Please, add a link to the demo so a reviewer can test your solution

koguts:
https://koguts.github.io/layout_landing-page/

PR #1743, opened 2024-10-24
create landing page

Andrii-Yanchuk:
https://andrii-yanchuk.github.io/layout_landing-page/


BudnikOleksii (changes requested):
Good work 👍 
A few changes are required in accordance to the design and the checklist:
image

- The button is too big
image

- The page shouldn't be reloaded on form submit, but the fields should be cleared. Use
image

- Disable page scrolling under the menu using the next code
image
image

BudnikOleksii (approved):
Good work 👍 
Approved, just one moment should be fixed:
image
image

PR #1747, opened 2024-10-24
new pull request

italomagno:


src/styles/main.scss
pedro-ruas, line 1: Italo, this file is waaay too big. Something with this amount of lines of code in a single file makes it really hard for maintenance.
I advise that you break it into smaller files, one file for each BEM Block you are creating the styles, and you can have another one for the base where you style tags in general.
This will make your code much more readable and easier for you.

pedro-ruas (changes requested):
Looks good, but with some things to change. I've added a comment there.
And, please remember to deploy your code using `npm run deploy` and add the links to the PR so we can see the website

src/styles/main.scss
italomagno: Thanks for you feedback @pedro-ruas I did some changes in my code. I think that stayed better to find where is each block.

src/layout/main/contacts.scss
```diff
@@ -0,0 +1,243 @@
...
+      font-weight: 700;
+      line-height: 27px;
+      text-align: center;
+      color: #fff;
+      border: none;
+      background-color: #D12D35;
+      cursor: pointer;
+
+
+      @media only screen and (min-width: 744px) and (max-width: 1024px) {
```
pedro-ruas, line 222: One great SCSS feature and advantage over simple CSS is the usage of Mixins.
This @media queries could be changed to @mixins instead, so you can avoid this long line repetition and make it more understandable.
For example:
```
@mixin on-tablet {
  @media only screen and (min-width: 744px) and (max-width: 1024px) {
    @content;
  }
}
```

Then, you can simply apply it wherever you need:
```
&_button {
  ...<button styling here>

  @include on-tablet {
      ...<media query conditional styling goes here>
   }
}
```

pedro-ruas: And this is valid for all media queries

pedro-ruas (changes requested):
Just a small change so it becomes perfect!

italomagno:
hey @pedro-ruas, now i having a serious problem... my command "sass ./src/styles/main.scss ./src/styles/main.css" is not working and giving the fallowing error:

Error: Undefined mixin.
   ╷
24 │ ┌     @include on-mobile {
25 │ │       margin-left: 20px;
26 │ │       font-family: Cinzel;
27 │ │       font-size: 40px;
28 │ │       font-weight: 700;
29 │ │       line-height: 32px;
30 │ │       letter-spacing: -1.5px;
31 │ │       text-align: center;
32 │ │
33 │ └     }
   ╵
  src\layout\header.scss 24:5  @forward
  src\styles\index.scss 3:1    @use
  src\styles\main.scss 1:1     root stylesheet

everything appears to be fine, but the code says it.

italomagno:
hey mr Pedro, i've solved the mixin problem writing it in each file. I hope that now, i can pass through this activity

joaorpereira (approved):

PR #1748, opened 2024-10-24
add task solution

schyrva:
[DEMO LINK](https://schyrva.github.io/layout_landing-page/)


BudnikOleksii (commented):
Good work 👍 
- Phone number should be centered
image

- Form looks different from the design

BudnikOleksii (changes requested):
.

BudnikOleksii (approved):
Good job 👍 
Approved, but I think the bold font in inputs is not necessary

PR #1754, opened 2024-10-25
add solution

Fridrif27:
[DEMO LINK] (https://fridrif27.github.io/layout_landing-page/)


BudnikOleksii (changes requested):
Good work 👍 
- Change **text** color on hover for phone, email, and address(you change background and it looks strange)
image

StasSokolov1 (approved):
Looks good to me, I'm approving it🔥

PR #1755, opened 2024-10-25
first commit

RuslanMudryi:
Цей проєкт є лендингом для сайту.
Технології
- HTML
- CSS
- SCSS
Посилання
Прев'ю: https://ruslanmudryi.github.io/layout_landing-page/
Макет: https://www.figma.com/design/NZQAIydtHo5QkINyGLHNcq/BIKE-New-Version?node-id=0-1&node-type=canvas&t=9HP47LVnvPCibFA7-0


BudnikOleksii (changes requested):
Good start 👍 
Pay attention to the design and the checklist:
image

- Make sure phone icons and phone numbers are real links to start a call. Email should be a valid mailto link as well
- Menu looks strange
image

image

- The textarea in the form looks different from the design
image

image

Moroz-Dmytro (changes requested):
1. Number is not centered
image

2. Add some hover effects when I hover over images(scale them for example)
3. When I click on address, it should open Google Maps in new tab
image

4. It looks strange on this width
image
image

5. I should be able to scroll this list
image

PR #1760, opened 2024-10-25
Landing My Bike

StricTGod:
[DEMO LINK](https://StricTGod.github.io/layout_landing-page/)

1. Use the logo as a favicon
3. All logos should be links to the top of the page
5. Make sure phone icons and phone numbers are real links to start a call
6. All addresses should be links to some location on Google Maps. Open them in a new tab with `target="_blank"`.
8. Make sure everything looks neat on mobile and without horizontal scrolling
9. The speed of animations should be the same throughout the page (for example, increasing when hovering or moving blocks when scrolling)
10. Ensure that all `nav__links` work smoothly
    ```css
    }
    ```
    ```css
    .menu {
    }
    .menu__top {
      top: 0;
      z-index: 1;
      background-color: #e5e5e5;
    }
    ```
13. All form fields should be required and have correct types (`email`, `tel`, etc.)
    ```html
    ```
16. Make the `footer` sticky at the bottom with `z-index: -1` to get a `fixed` background effect.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -18,7 +18,7 @@
     <link
       rel="preconnect"
       href="https://fonts.gstatic.com"
-      crossorigin
+      crossorigin="anonymous"
     />
     <link
       href="https://fonts.googleapis.com/css2?family=Poppins:wght@400;500;600;700&family=Roboto:wght@500&display=swap"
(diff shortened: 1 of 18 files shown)


BudnikOleksii (changes requested):
Good start 👍 
Please, implement all requirements from the checklist that you have copied to the PR description. 
Also, a few moments should be fixed in accordance with design:
image

- Inputs don't have paddings and font are different
image

- Footer looks strange
image

PR #1766, opened 2024-10-26
add task solution

naz4ik:
Hello my first landing!

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
+/backstop_data
 /dist
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,3 @@
 module.exports = {
-  extends: ['@mate-academy/eslint-config', 'plugin:cypress/recommended'],
+  extends: '@mate-academy/eslint-config',
 };
--- a/.gitignore
+++ b/.gitignore
@@ -9,10 +9,7 @@ node_modules
 .DS_Store
 
 # Generated files
+backstop_data
 dist
 .cache
 .parcel-cache
-raw_reports
-cypress/screenshots
-cypress/videos
-reports
--- a/.stylelintignore
+++ b/.stylelintignore
@@ -1,1 +1,2 @@
+/backstop_data/
 /dist
--- a/README.md
+++ b/README.md
@@ -4,17 +4,14 @@
 > Now just choose a design and do the parts from the videos
 
 Choose a design for your landing page:
-- [MET landing](https://www.figma.com/file/lSR1m42L9YwzQwzzxKwHpw/THE-MET)
-- [MYBIKE landing](https://www.figma.com/file/NZQAIydtHo5QkINyGLHNcq/BIKE-New-Version?node-id=0%3A1)
-- [B&O](https://www.figma.com/file/DtkQmQ797hk0nI4KfMi2Uq/BOSE-New-Version?type=design&node-id=6817-212&t=ZTV6Gl8NzaWkJ4FK-0)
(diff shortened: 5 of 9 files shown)


TarasHoliuk (changes requested):
It seems like you haven't finished the task yet (I see no styles in scss file)

Also, do not forget to deploy your page when the task will be done:

image

And add a link to the deployed page to PR description (same as for other layout tasks) - point 19: https://mate-academy.github.io/layout_task-guideline/

TarasHoliuk:
1. You should set `no-repeat` for [background-repeat](https://developer.mozilla.org/en-US/docs/Web/CSS/background-repeat) property:
image

2. You add a border to inputs by default but remove it on focus (no need to remove it, at least you can change its color to show the input is focused). It changes the input's size and surround content "jumps" on focus and blur (blur is when you unfocus the element)

image

3. It should be links. Also, add a hover effect (to all interactive elements):

image

4. On the phone icon hover you should add a phone link:
image

5. Read the [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md). Make sure you've reached all points (1, 4, 5, 6, 7, 11, 12, 13, 15, 16)

TarasHoliuk:
P.S. Push all files to GitHub (as discussed in the chat)

TarasHoliuk (changes requested):
Hey! You did a good job! But let's improve your landing page a bit 🙂

TarasHoliuk:
1. Checklist point 1 - add favicon

2. Do not set background-size: 100%:
image

Without this property image looks as expected:
image

It's okay if it's cropped on some screens. In such cases, you should set the correct background-position to show the correct part of the image. But in this case 'center' is fine - no need to change background-position.

3. Now transition works only in one way. To make it work in both ways add the transition to the element itself, not to the element:hover.
image

Also, you shouldn't create a separate BEM element `detail__hover`. Just use the current element and `:hover` pseudoclass

4. Do not add a hover effect and link to the description:

image

You should add it only to the actual links (below each description).

5. All text below `Address` (I.e. map link) should be one element instead of 3 separate elements with own hover effect each:
image

6. Phone should be a phone link, email - email link with mailto, and address should be links to some map (for example google map) with its address:

image

Do not forget to open the map link in the new tab (it's a general rule for all external links).

7. The title in the menu looks bad (wrong background color). Scroll is still enabled when the menu is open. Again - read the checklist and make sure all points are reached.

8. All fields in the form should be required (read instructions about the form in the checklist)

src/styles/details.scss
```diff
@@ -0,0 +1,98 @@
...
+}
+
+.detail {
+  &__photos {
+    @include page-grid;
+
+    margin-bottom: 16px;
+  }
+
+  &__hover {
```
TarasHoliuk, line 19: such &__hover elements are redundant, use just `&:hover`

src/styles/footer.scss
```diff
@@ -0,0 +1,17 @@
+.footer {
+  content: '';
```
TarasHoliuk, line 2:
```suggestion
```
redundant

Footer should be sticky (checklist 16)

VitaliyBondarenko1982 (changes requested):
Good job. Almost done.
- Nothing happened after click on send button. 
  Please pay attention on [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md).
  Point 15. The page shouldn't be reloaded on form submit, but the fields should be cleared.
All form elements should be in tag form.

anastasiiavorobiova (changes requested):
Great work!
- Screenshot 2024-11-04 at 17 34 54
- Check the alignment and margins here

etojeDenys (changes requested):
1. disable the scroll if the menu is open
image

2. looks good

anastasiiavorobiova (approved):
Awesome!

PR #1774, opened 2024-10-28
Develop

pogorielova:
[DEMO LINK](https://pogorielova.github.io/layout_landing-page/)


BudnikOleksii (changes requested):
Good work 👍 
A few moments that should be fixed:
- When a user hovers over the phone icon you should show the number
image

- The content in the footer and in other sections should have the same max-width
image

- All interactive elements such as buttons should have a cursor pointer
- I'm not sure about increasing buttons on hover, they looks too big, but it's up to you

StasSokolov1 (changes requested):
Hey, a few comments left to fix before approval!

1. User can resize textarea like this, need to restrict this possibility
Screenshot 2024-10-31 at 11 38 25

2. Let's add cursor pointer for recommended section
Screenshot 2024-10-31 at 11 38 15

BudnikOleksii (approved):
LGTM 🚀

PR #1777, opened 2024-10-28
add task solution

thshnhta:
[DEMO LINK](https://thshnhta.github.io/layout_landing-page/)


BudnikOleksii (changes requested):
Please, implement the landing page in accordance with a
image
 design at first and pay attention to the [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md):
- Use the logo as a favicon
- When a user hovers over the phone icon you should show the number
image
- Make sure phone icons and phone numbers are real links to start a call. Email should be a mailto as well
- The speed of animations should be the same throughout the page (for example, increasing when hovering or moving blocks when scrolling)
- Make sure everything looks neat on mobile and without horizontal scrolling
image
image
image

image
- Form doesn't work
image

PR #1781, opened 2024-10-29
add solution

save-v:


BudnikOleksii (changes requested):
Good work 👍 
Just a few changes are required:
- When a user hovers phone icon you should show the number
image

- Menu looks strange
image

- All interactive elements such as buttons should have a cursor pointer

VolodymyrKirichenko (approved):
Well done 🔥

PR #1786, opened 2024-10-31
add task solution

RomanHasiuk:
- DEMO LINK](https://RomanHasiuk.github.io/layout_landing-page/


SanyaBratashchuk (changes requested):
1. Disable scroll when menu is open 
2. get rid of horizontal scroll
3. form shouldn't trigger page reload after submit

etojeDenys (changes requested):
1. number should appear on hover
image

2. should be the links
image

3. button must have submit type
image

4. add some favicon
image

5. don't forget to follow this file
https://github.com/RomanHasiuk/layout_landing-page/blob/develop/checklist.md

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Change title on the page to `B&O` 
image

2. You need to make a call when user clicking here 
image

3. Add a some hover effects for images 
image

4. These elements must be links 
image

5. Add hover effects here 
image

image

7. Don't need to use position for the footer 
image

8. You need to remove the horizontal scroll on the mobile version 
image

anastasiiavorobiova (changes requested):
Great work! Almost done!
- Screenshot 2024-11-02 at 16 50 02
- The image shouldn't overlap the title on the hover
- Screenshot 2024-11-02 at 16 49 51
- Check hovers here
- Consider using the `mailto` lint and the link to Google Maps

RomanHasiuk:
I can't figure out what I should do because I don't see any problem with this
Знімок екрана 1

RomanHasiuk:
What's wrong? What should I do? I don't see like that!

TarasHoliuk:
This issue is still not fixed:

image

Everything else looks great, so I'll approve this task 👍 I'd recommend you fixing this issue before adding the landing page to the portfolio (there'll be a separate task for it later)

TarasHoliuk (approved):
Approved with one comment

PR #1789, opened 2024-11-03
add solution

t-namestnykova:
- [DEMO LINK]   https://t-namestnykova.github.io/layout_landing-page/


SanyaBratashchuk (changes requested):
1. Disable page scroll when menu is open 
image

vadiimvooo (changes requested):
Good job, a few changes required
1. Image should not overlap the neighbours on hover
image
2. Add transition on hover effect
image
3. Add hover effect for inputs
image
4. Remove scroll from textarea
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add transition when the menu appears on the page 
image

2. You don't need to open page in the new tab after click on the menu link 
image

image

PR #1801, opened 2024-11-06
add task solution

kovaltar:
[DEMO LINK](https://kovaltar.github.io/layout_landing-page/)


SanyaBratashchuk (changes requested):
image
2. should be links 
image
3. page shouldn't be refresh after form submit

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. Hide the additional phone number on the mobile and tablet versions 
image

2. The font family must be are same everywhere 
image

3. Fix the background image, check it on the design. The top cyclist must not be clipped
image

etojeDenys (approved):
lgtm

PR #1808, opened 2024-11-10
layout_landing-page

TaniaFedortsi:


VitaliyBondarenko1982 (changes requested):
Please provide DEMO LINK to pr description

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. You need to show additional phone number when hovering this icon and make a call when user clicks here 
image

2. Add hover effects for images 
image

3. Fix height for the textarea 
image

4. The user cannot be able to change size of the textarea field 
image

5. Add cursor pointer for this button 
image

image

7. These elements must be links 
image

8. The height for the menu must be 100vh 
image

etojeDenys (changes requested):
1. 
 image
number should appear on hover

2. disable textarea resizing
image


3. onclick should start the call
image

4. add cursor: pointer to button
image

volodymyr-soltys97 (changes requested):
Hi, the comments from the previous review still not fixed, you need to make deploy after your changes and check it after on the demo link

maxim2310 (changes requested):
Almost done
1. cursor pointer should be here also 
image
2. add some space or decrease scale on hover 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Hover effects should be for each element separately.
image

2. The border radius looks different as on the design, need to fix 
image

3. The color should be are same everywhere 
image

4. Remove default blue border when input is focused 
image

VitaliyBondarenko1982 (approved):
Good job.

PR #1810, opened 2024-11-10
landing page done

mtereshyn:


volodymyr-soltys97 (changes requested):
Hi, there is no demo link, also, you need to add demo link to the PR description 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Fix styles for the phone number, check it on the design 
image

2. Need to remove scroll when the menu is open 
image

image

4. Fix default background color for this button and add cursor pointer here 
image

image

6. Add hover effects here 
image

7. Hide the additional phone number on the mobile and tablet versions 
image

mtereshyn:
2. Need to remove scroll when the menu is open

Can you help me with this issue?

PR #1828, opened 2024-11-22
add task solution

troleksii:
[DEMO LINK](https://troleksii.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. disable textarea resizing
image

2. add some favicon
image

3. add hover for all links
image

4. number should appear on hover
image

5. disable the scrolling if the menu is open

vadiimvooo (changes requested):
Good job, there are some fixes to do:
1. It should be possible to click the number as well (at the moment it's not possible to even hover it)
image
2, Transition effect should be when you remove cursor from element as well.
3. Elements on hover effect should not overlap neighbor elements.
image
4. Remove default autofill styles
image

anastasiiavorobiova (approved):
Great work!

PR #1833, opened 2024-11-22
add task solution

Olifan:
[DEMO LINK](https://olifan.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add the topical favicon on the page 
image

2. Check size for the logo on the design 
image

3. You need to show additional phone number when hovering this icon 
image

4. Need to disable scroll when the menu is open 
image

image

6. The user cannot be able to change size of the textarea field and fix border radius
image

image

8. Add cursor pointer for all clickable elements 
image

9. All these elements must be links and add hover effects here 
image

VitaliyBondarenko1982 (changes requested):
Good job.
Almost done

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Change title on the page to `My Bike` 
image

2. The prices must be are on the same horizontal line and check font weight here
image

3. Add hover effects for these images 
image

image

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. Fix font size for the menu links on the mobile version, check it on the design 
image

2. Remove padding bottom and margin here on the mobile versions 
image
image

volodymyr-soltys97 (approved):
Great work! 🔥

PR #1835, opened 2024-11-23
Develop

staner1:
https://staner1.github.io/layout_landing-page/


anastasiiavorobiova (changes requested):
Great work!
- Screenshot 2024-11-23 at 16 57 17
- Check styles here
- Screenshot 2024-11-23 at 16 56 45

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Remove position for the header 
image

image

image

image

image

PR #1851, opened 2024-12-02
Add task Solution

MannyShussan:
[DEMO LINK](https://MannyShussan.github.io/layout_landing-page/)


julianFBarbosa:
image
demo is unavailable

src/styles/blocks/icon.scss
```diff
@@ -0,0 +1,61 @@
...
+      content: '+1 234 5555-55-5';
+      font-size: 16px;
+      font-weight: 700;
+      line-height: 32px;
+      color: #000;
+      position: absolute;
+      width: 140px;
+      right: 50px;
+
+      transition: transform 0.5s ease;
```
julianFBarbosa, line 38: transition doesn't has the same speed as the rest of the others

image

src/styles/blocks/menu.scss
```diff
@@ -0,0 +1,41 @@
...
+.menu {
+  height: 100vh;
+  box-sizing: border-box;
+  background-color: #f9f9f9;
+  font-weight: 700;
+  font-size: 16px;
+  line-height: 22px;
+
+  &__top {
+    @include menu-top-margin-bottom;
```
julianFBarbosa, line 12: image

I don't see any styles that makes it possible

julianFBarbosa (changes requested):
Left some comments that i could see without a demo link, I will review it again after you update/re-deploy your landing page

pedro-ruas (changes requested):
Your works looks very good, and the website works mostly great.
However, you did not fix one of the previous requests from the other mentors, and the Menu is not scrollable when it needs to be. Fixing that, will make it perfect

PR #1858, opened 2024-12-05
Develop

maksym-korchynskyi:
# Landing page

[DEMO LINK](https://maksym2493.github.io/layout_landing-page/)

# Description

During execution, the BEM methodology was used.

To build the header of the site, the semantic tags header and nav were used. For the content: main, section and article. For the bottom of the site: footer, form, section, article.

To adapt to different devices, media queries were used.

For styling, the SASS preprocessor was used: creating changes, mixins and working with nested classes.

Also used were such display methods as: flex, grid, and various ways of positioning elements in blocks: relative and absolute.

In addition, for smooth changes, transition-duration was set, and to change the elements themselves, transform properties were used.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -374,7 +374,10 @@
             <h2 class="section-title contact__title">Contact us</h2>
 
             <div class="contact__content">
-              <form class="contact-form contact__form">
+              <form
+                class="contact-form contact__form"
+                onsubmit="this.reset(); return false;"
+              >
                 <input
                   type="text"
                   name="name"
@@ -405,7 +408,7 @@
                 ></textarea>
 
                 <input
-                  type="reset"
+                  type="submit"
                   value="Send"
                   class="primary-button contact-form__field--submit-button"
                 />


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add the topical title on the page 
image

2. Fix size for these icons, check it on the design 
image

3. Remove scroll when the menu is open 
image

4. The prices must be are on the same horizontal line 
image

5. Change the autofill styles when the menu is open
image

6. After submit the form don't need to refresh page and scroll to top, you need only reset form fields 
image

7. Add a correct address on google maps here 
image

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. Need to open this additional link in the new tab 
image

etojeDenys (approved):
should be the button with type submit
image

and don't forget to handle the form submitting
image

maksym2493:
I did it. Thank you.

PR #1860, opened 2024-12-06
finished the task

mykhailokurochkin:
- [DEMO LINK](https://Mihakurochkin.github.io/layout_landing-page/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -53,10 +53,14 @@
               />
             </a>
             <div class="top-bar__icons">
-              <a
-                href="tel:+1 234 5555-55-55"
-                class="icon icon--phone"
-              ></a>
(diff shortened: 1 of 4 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Fix size for these icons, check it on the design 
image

2. You need to show additional phone number when hovering this icon 
image

3. Add transition when this menu appears on the page and remove scroll when menu is open 
image

4. Fix font weight for the prices 
image

5. Add transition for hover effects everywhere 
image

6. Fix gap between these elements 
image

VitaliyBondarenko1982 (approved):
Looks good

PR #1887, opened 2024-12-19
add task solution

GumenchukAN:
-[DEMO LINK](https://GumenchukAN.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job! 👍 
Let's improve your page
image

2. Add the topical title and favicon on the page 
image

3. Remove scroll when the menu is open 
image

4. You need to change image for header, without red border 
image

image

6. Fix gap between these elements, check it on the design 
image

image

8. Add cursor pointer for all clickable elements 
image

9. All these elements must be links and add hover effects here 
image

etojeDenys (changes requested):
do not open new tab on phone click
image

2. button should be inside the form
3. add "required" attribute to all inputs
image

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. Change title of the page to `B&O`. Title should be actual
image

2. Fix styles and position for the phone number, check it on the design. Also, you need to make a call when user clicks here
image

image

4. Add hover effects for these blocks 
image

5. After submit the form don't refresh page and scroll to top, you need only reset form fields 

6. Add the topical address on google maps for this link 
image

7. This element should be link with `mailto:` 
image
image

8. The header looks strange on the mobile, fix it 
image

9. Also, you need to hide the additional phone number on the mobile and tablet versions 
image

etojeDenys (changes requested):
1. button type should be submit
image

2. 
image


3. you should stretch this bg
image

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
image

VitaliyBondarenko1982 (approved):
Looks good

PR #1889, opened 2024-12-20
add task solution

GabrielNunes-383:
[DEMO LINK](https://GabrielNunes-383.github.io/layout_landing-page/)


pedro-ruas (changes requested):
Gabriel, code-wise it looks ok, but there are a few issues that you missed:
- I did not find any animation in the page, the images t=should be animated on hover and the same is valid for the Phone at the top of page
- Most links do not work, it should be possible to navigate the page from the links in it, but most of them don't do anything
- Contact phone, email and map links beside the Contact Form should also work and open outside interactions
- Contact Form should only be able to be sent if complete, and when sent it should NOT reload the page, neither send you to another page

raulriato (approved):
As the changes suggested by Pedro were implemented, I'll approve it.

PR #1891, opened 2024-12-22
Develop

ArtemHryb:
- [DEMO LINK](https://ArtemHryb.github.io/layout_landing-page/).


volodymyr-soltys97 (changes requested):
Good job!
Let's improve your page
1. You need show additional phone number when hovering this icon, check it on the design and add hover effects here 
image

2. The images shouldn't overlap each other when hovering images 
image

image

image

5. Add cursor pointer for all clickable elements 
image

6. Add hover effects here 
image

7. Fix height for the footer, check it on the design 
image

etojeDenys (approved):
lgtm

PR #1894, opened 2024-12-23
add solution

VM93:
- [DEMO LINK](https://VM93.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good start!
To improve:
1. You need to fix background image in the header 

2. Keep going with next sections

If you need some help pls feel free in the fe_chat

PR #1908, opened 2024-12-31
add task solution

NadiiaBulmak:
- Deploy and create a Pull Request with a [DEMO LINK](https://NadiiaBulmak.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job!
To improve:
image

2. The image here should stretch to the full width of the browser window.
image

3. Fix the form, check it in the design 
image

4. Change the autofill styles for the form fields
image

image

6. After submit the form don't need refresh page and scroll to top, you need only reset form fields

image

NadiiaBulmak:
Hi,
Should I use JS for improving the mistake about 6)After submit the form don't need refresh page and scroll to top, you need only reset form fields)?

VitaliyBondarenko1982 (changes requested):
Good job.
Almost done.
Prevent scroll page when menu is open.
Screenshot 2025-01-02 at 15 00 56

etojeDenys (approved):
lgtm

PR #1910, opened 2024-12-31
add task solution

nikol-nuzhna:
[DEMO LINK](https://nikol-nuzhna.github.io/layout_landing-page/).


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
image

2. Remove scroll when the menu is open 
image

3. This section looks different as on the design 
image

Design 
image

4. Fix the max width for content, it should be as on the design 
image

5. Where is the text in this button? 
image

6. Change the autofill styles for the form field
image

7. After submit the form don't need refresh page and scroll to top, you need only reset form fields 
image

8. Add hover effects here 
image

9. This element should be link 
image

nikol-nuzhna:
Unfortunately, I haven't found a way to submit form without refreshing the page using only html/css. The other comments have been fixed.

maxim2310 (changes requested):
probably you forgot to deploy your fix, also if you have questions concerning comments feel free to ask in chat

PR #1914, opened 2025-01-03
My LandingPage

TaniaP26:
https://TaniaP26.github.io/layout_landing-page


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add a favicon on the page and change title to `The MET` 
image

2. Add hover effects for all clickable elements on the page 
image

3. You need to show additional block with phone number when hovering this icon and make a call when user clicks here 
image

4. Add transition when this menu appears on the page and remove scroll when menu is open 
image

5. The images shouldn't overlap each other when hovering 
image

6. These elements should be links and add hover effects here 
image

TaniaP26:
Thanks for comments. I fixed everything almost but one. I don't know how  remove scroll when menu is open because this scroll belongs to the page__body class. I just added  z-index=100 to the menu.

anastasiiavorobiova (approved):
Great work!

TaniaP26:
Thanks!

PR #1930, opened 2025-01-09
add task solution

VictorKomara:
https://VictorKomara.github.io/layout_landing-page/


volodymyr-soltys97 (changes requested):
Good job! 👍 
Let's improve your page
1. Add a favicon on the page 
image

2. You need to show additional phone number when hovering this icon and make a call when user clicks here 
image

3. The left indent should be same 
image

4. Need to remove scroll when the menu is open 
image

image

image

7. Fix height for these fields, check it on the design 
image

image

9. These elements should be links and add hover effects here 
image

etojeDenys (approved):
good job

PR #1948, opened 2025-01-14
add solution

ratseburzhynskyi:
[DEMO LINK](https://ratseburzhynskyi.github.io/layout_landing-page/)


ratseburzhynskyi:
Thanks, I'll do it. But the checklist says this background is needed.

So, I can remove it, and also remove 'position: sticky'. Or I can set another color, e.g. #f9f9f9 (as the menu background color), and add a shadow to separate top bar and navigation in case of scrolling.

What solution will be better?
Screenshot 2025-01-14 164907

PR #1951, opened 2025-01-14
Add project

AlinaYamchuk:
[DEMO LINK](https://AlinaYamchuk.github.io/layout_landing-page/)


vadiimvooo (changes requested):
Good job, there are a few changes to fix:
1. Elements on hover should not overlap neighboor elements
image
2. Check and fix letter spacing for this title
image
3. Rework the section below according to mockup
image
image
4. The same for Contact Us section
5. Fix height of textarea and remove possibility to resize it.
image
6. Use for email input proper type tag
7. Remove autocomplete styles
image
8. Send button should not reload the page. Instead just clear the fields.
9. Add scroll if menu content height is greater than height of viewport
image
10. Fix position of the sections on mobiles
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Fix gap between these elements 
image

2. All images on the page should scale down as smoothly as they scale up on hover.
image

3. Why did you remove autocomplete for the form fields?
You only need to change the autocomplete styles.
image

image

5. You need to reset form fields after submit the form 
image

6. Check gap between these block on the mobile version 
image

VitaliyBondarenko1982 (changes requested):
Good job.
Almost done.

- make contact us section width the same as other content.
Screenshot 2025-01-15 at 15 27 23

-  after click on send button nothing happens. Form should be cleared but page shouldn't be reloaded. For it use 
and send button should be inside form tag with type submit.

volodymyr-soltys97 (changes requested):
Hi, these comments still not fixed from the previous. review, if you need some help feel free in the fe_chat 
image

maxim2310 (changes requested):
I steel can submit form even fields are not filled in
image
your button should be submit type
image

PR #1952, opened 2025-01-14
add task solution

prutasveta:


sarakhmen (changes requested):
the task was assigned to the wrong mentors group. It's a frontend task but got to java mentors. Please request an additional review and let's see if the proper mentors group is assigned

PR #1956, opened 2025-01-14
add solution

KOliinyk:
- [DEMO LINK](https://KOliinyk.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. The background image should stretch to the full width of the browser window and the content should be centered.
image

2. Add the topical title on the page 
image

3. You need to show additional phone number when hovering this icon and make a call when user clicks here 
image

4. Remove scroll when the menu is open 
image

5. Add hover effects here 
image

6. Add transition for hover effects everywhere

7. The font family should be are same everywhere 
image

8. These elements should be links and add hover effects here 
image

etojeDenys (changes requested):
disable the scroll if menu is open
image


2. don't forget to follow these steps
image

3. fix these mismatches
image

etojeDenys (approved):
add the `required` attribute to all inputs
image

PR #1959, opened 2025-01-15
add task solution

grytchyn:
https://grytchyn.github.io/layout_landing-page/


vadiimvooo (changes requested):
Good job, keep going and finish other parts of task.
1. On phone icon hover phone number should appear
image
2. Remove scroll when menu is opened
image

PR #1960, opened 2025-01-15
add task solution

RHerheman:
[DEMO LINK](https://RHerheman.github.io/layout_landing-page/).


volodymyr-soltys97 (changes requested):
Good job! 👍 
To improve: 
1. Add a favicon on the page 
image

image

3. Check font size for the phone number 
image

4. The prices should be on the same horizontal line 
image

image

6. Add cursor pointer for this button 
image

7. Change the autofill styles for the. form fields
image

8.  Add hover effects for these links 
image

9. This element should be link with `mailto` 
image

PR #1966, opened 2025-01-16
Nothing landing page

barbanihor:
https://www.figma.com/file/DtkQmQ797hk0nI4KfMi2Uq/BOSE-New-Version?type=design&node-id=6802-139&t=L7eKz5YKLN0m5WxR-0

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
+/backstop_data
 /dist
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,3 @@
 module.exports = {
-  extends: ['@mate-academy/eslint-config', 'plugin:cypress/recommended'],
+  extends: '@mate-academy/eslint-config',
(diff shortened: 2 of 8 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. The image is currently taking a very long time to load because it is over 6MB, with a bad internet connection it will take even longer to load, try optimizing it
image

2. Add a favicon on the page 
image

image

image

5. Don't need to use fixed width for the content on the mobile and tablet, use 100% instead 
image

vadiimvooo (changes requested):
Good job, a few things to improve:
1. It's not possible to hover the number, please fix it
image
2. Add favicon
image
3. Fix font-family in textarea
image
4. Field hover should not move text inside
image

etojeDenys (approved):
looks good, just don't forget to add the favicon

PR #1970, opened 2025-01-17
Landing page Bike

andrei-pcsantos:


package.json
```diff
@@ -1,58 +1,12 @@
 {
-  "name": "layout_landing-page",
-  "version": "1.0.0",
-  "description": "Landing page",
+  "name": "landing",
+  "source": "src/index.html",
   "scripts": {
```
joaorpereira, line 4: why you removed the scripts? This is necessary.

The tests are failing and it's missing the DEMO url.

joaorpereira (changes requested):

PR #1975, opened 2025-01-19
page is implemented

SamarashEugene:
https://samarasheugene.github.io/layout_landing-page/


vadiimvooo (changes requested):
Good job, there are a few comments to fix:
1. Phone number should not dissapear on hover
image
2. All sections should have the same limits
image
3. Hovered elemets should not overlap neighboor elements. Please fix it.
image
4. Send button should reset input fields.
5. Add type email for email input.
6. Add scroll in menu if content height is greater than viewport.

volodymyr-soltys97 (changes requested):
Good job!
To improve:
image

2. Remove default blue border when input is focused 
image

3. Add hover effects for these links 
image

image

5. Need to center the background image 
image

volodymyr-soltys97 (approved):
Good job! 👍

PR #1982, opened 2025-01-22
Develop

zhuraveldasha:

-Used BEM and SCSS

https://zhuraveldasha.github.io/layout_landing-page/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -317,7 +317,7 @@
                 <div class="contacts__input">
                   <button
                     class="contacts__input--button"
-                    type="submit"
+                    type="reset"
                   >
                     Send
                   </button>
--- a/src/styles/blocks/categories.scss
+++ b/src/styles/blocks/categories.scss
@@ -17,6 +17,7 @@
     grid-column: span 1;
 
     &:hover {
+      transition: transform 0.3s;
       transform: translateZ(10px);
     }
 
--- a/src/styles/blocks/contacts.scss
+++ b/src/styles/blocks/contacts.scss
@@ -24,6 +24,7 @@
       margin-bottom: 24px;
 
       &:hover {
+        transition: transform 0.3s;
         font-weight: 700;
       }
     }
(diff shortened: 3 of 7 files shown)


volodymyr-soltys97 (changes requested):
Hi, your demo link isn't working, you need to make a deploy and check the demo link after 
image

volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. Fix size for these icons on the tablet 
image

2. Fix styles for the phone number and you need to make a call when user clicks here 
image

3. Add hover effects for all images 
image
image

image

5. Add cursor pointer for this button 
image

6. All these elements should be links and add hover effects here 
image

zhuraveldasha:
1.Fix size for these icons on the tablet, what is wrong with the size it is 32px as it should be?
2.Fix styles for the phone number and you need to make a call when user clicks here.
How can it be implemented without using scss? I solved this task using a pseudo-class and pseudo-element &:hover::before

zhuraveldasha:
1 .Fix size for these icons on the tablet. what's wrong with the size?
2.Fix styles for the phone number and you need to make a call when user clicks here.
I solved this problem using &:hover ::before , how can this be done without using js?
6. All these elements should be links and add hover effects here. What exactly should :hover do?

vadiimvooo (changes requested):
Previous comments should be fixed. Please feel free to ask for help in fe_chat.

volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. You need to add transition for all hover effects on the page 

2. After submit the form don't need refresh page and scroll to top, you need only reset form fields 
image

3. Need to add additional styles for inputs in different states
image

etojeDenys (changes requested):
disable the scroll if menu opened
image

2. All addresses should be links to some location on Google Maps. Open them in a new tab with target="_blank".
3. fix small border jumps on inputs hover
4. add the `required` attribute to all inputs

VitaliyBondarenko1982 (approved):
Good job.

During prepare project for portfolio recommend to check all elements with hover effect.
Transition should be on all elements with hover effect. And should work in both ways (add transition for .selector instead .selector:hover)

PR #1998, opened 2025-01-28
Develop

osidanych:
https://osidanych.github.io/layout_landing-page/


vadiimvooo (changes requested):
Good job, there are a few changes to fix:
1. It should be possible to hover the number as well
image
2. Images on hover should not overlap neigboor elements
image
3. Image should take 100% width
image
4. Menu should also have background 100% widht
image

osidanych:
> Good job, there are a few changes to fix:
> 
> 1. It should be possible to hover the number as well
>    image
> 2. Images on hover should not overlap neigboor elements
>    image
> 3. Image should take 100% width
>    image
> 4. Menu should also have background 100% widht
>    image

osidanych:
> 2. Images on hover should not overlap neigboor elements
Sorry, but I can't see that images overlap neighbor elements. 
image

PR #2000, opened 2025-01-28
The MET

mykolahuryn:
[DEMO LINK](https://mykolahuryn.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add hover effects for all clickable elements on the page 
image

2. Fix the left padding here 
image

3. The mobile version looks strange 
image

4. You need to make a call when user clicks here 
image

volodymyr-soltys97 (changes requested):
Hi, you need to fix height and width for inputs, check it on the design 
image

volodymyr-soltys97 (approved):
Good job! 👍

PR #2008, opened 2025-01-31
header i aside

Diana2369:
[DEMO LINK](https://Diana2369.github.io/layout_landing-page/)


etojeDenys (changes requested):
good job, you will finish this task later so move forward for now

volodymyr-soltys97 (changes requested):
Hi, there is no styles on the demo link, you need to make a deploy and check after the demo 
image

etojeDenys (changes requested):
1. disable the scroll if menu is open
image


2. show number on hover
image


3. should be links
image


4. don't forget to follow this file https://github.com/Diana2369/layout_landing-page/blob/develop/checklist.md

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Need to center the phone number, check it on the design 
image

image

image

image

5. Add the correct address on maps for this link 
image

6. Hide the additional phone number on the mobile and tablet versions 
image

7. You need to make a call when user clicking 
image

vadiimvooo (changes requested):
Good job, there are a few fixes required. Please feel free to ask for help in fe_chat.
1. Add transition on all hover effects (for example this button)
image
2. It should be possible to hover the number as well.
image
3. Add scroll if elements height is greater then viewport height
image

VitaliyBondarenko1982 (approved):
Looks good

PR #2023, opened 2025-02-04
Develop

Vasylevsk:
[DEMO LINK](https://Vasylevsk.github.io/layout_landing-page/).


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
image

2. Unable to hover over phone number
image

image

4. Fix height for the textarea 
image

5. These elements should be links 
image

PR #2026, opened 2025-02-05
Develop

saywin:
[DEMO LINK](https://saywin.github.io/layout_landing-page/).


maxim2310 (changes requested):

1. number should be clickable 
image
2. prices should be on same level 
image

VitaliyBondarenko1982 (approved):
Looks good

PR #2029, opened 2025-02-05
Develop

DmytroHoncharuk:
[DEMO LINK](https://dmytroHoncharuk.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add a favicon on the page 
image

2. Remove scroll when the menu is open 
image

3. Add hover effects for these blocks and images 
image
image

4. After submit the form don't need refresh page and scroll to top, you need only reset form fields 
image

5. Add hover effects for these links 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. This comment still not fixed 
image

2. Hover effects should be for each image separately. 
image

3. The font family should be are same 
image

4. Change the autofill styles, the autofill and default styles should be are same 
image

6. Fix height for the textarea, check it on the design 
image

7. Not fixed 
image

PR #2035, opened 2025-02-06
Createt landing for The MET

daniil-trusov:
- [DEMO LINK](https://daniil-trusov.github.io/layout_landing-page/)
- [CODE REPO LINK](https://github.com/daniil-trusov/layout_landing-page/tree/dev)

[x] Use the logo as a favicon
[x] Change text color on hover for phone, email, and address (I implement another solution - underline on hover)
[x] All addresses should be links to some location on Google Maps. Open them in a new tab with target="_blank".
[x] Increase all images on hover (make them links to #, if there is no better option)
[x] Ensure that all nav__links work smoothly
[x] Disable page scrolling under the menu
[x] All form fields should be required and have correct types (email, tel, etc.)
[x] Placeholders should give examples of what should be added, to make the expected format clear
[x] The page shouldn't be reloaded on form submit, but the fields should be cleared.
[x] Make the footer sticky at the bottom with z-index: -1 to get a fixed background effect.


alexander-ignatow (approved):
Good job, but some issues are here:

image
If I am not mistaking, you need to show the numbers on hover on this phone icon

Good job in all the other, so let's move on with this

PR #2046, opened 2025-02-07
The Met

sansamiste:
[DEMO LINK] [https://sansamiste.github.io/layout_landing-page/]


alexander-ignatow (changes requested):
Good job overal

However, you somewhat misuse BEM and need to add some fixes to your project structure

The intent of BLOCK in BEM that blocks are reusable. If you have the block with 2-3 lines of styles - that should not be a block. If you have two blocks that are only used together - please combine them into one

You need to reorganise your project structure and reduce the number of blocks (for instance you don't need body_nav-menu and nav-menu separately)

And one minor visual issue: 
Probably you should add transform on hover on each picture separately, not on the whole section
image

alexander-ignatow (approved):
Looks good now!

PR #2047, opened 2025-02-07
Develop

PiekharievaK:
-[DEMO LINK]  https://PiekharievaK.github.io/layout_landing-page/


src/index.html
```diff
@@ -16,7 +37,389 @@
...
+              class="contacts__form-field"
+            >
+              <input
+                type="text"
+                placeholder="Name: John Dee"
+                class="contacts__input"
+                required
+              />
+              <input
+                type="text"
```
alexander-ignatow, line 369:
```suggestion
                type="email"
```

src/index.html
```diff
@@ -16,7 +37,389 @@
...
+                required
+              />
+              <textarea
+                name="message"
+                placeholder="Message:"
+                rows="4"
+                class="contacts__input contacts__input--textarea"
+                required
+              ></textarea>
+              <button class="contacts__button button">Send</button>
```
alexander-ignatow, line 381: It makes more sense using submit input, if you are using the form like it is, without any additional JS processing
```suggestion
              <input type="submit" value="Send"  class="contacts__button button"/>
```

src/styles/blocks/button.scss
```diff
@@ -0,0 +1,27 @@
+@use '../variables.scss' as *;
+
+.button {
+  display: flex;
+  box-sizing: border-box;
+  width: 100%;
```
alexander-ignatow, line 6: minor - but avoid including width and margin  on BEM block level as it reduces reusability

src/styles/blocks/compare.scss
```diff
@@ -0,0 +1,16 @@
+@use '../utils/mixins.scss' as *;
+
+.compare {
+  margin-bottom: 92px;
```
alexander-ignatow, line 4: no margins on blocks

src/styles/blocks/icon.scss
```diff
@@ -0,0 +1,85 @@
...
+  @include on-tablet {
+    width: $--big-icon-size;
+    height: $--big-icon-size;
+    &:not(:last-child) {
+      margin-right: $--big-icon-size;
+    }
+  }
+}
+
+.icon--phone {
```
alexander-ignatow, line 26: You can nest this withtin `.icon`

src/styles/blocks/menu.scss
alexander-ignatow, line 1: Why no nesting hin this file?

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,141 @@
...
+  margin: 0;
+}
+
+a {
+  text-decoration: none;
+}
+
+button {
+  border: none;
+}
```
alexander-ignatow, line 14: This should go to normalize or reset file
At the very least move it to some separate file, as it has nothing to do with `page` block

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,141 @@
...
+    margin-bottom: 32px;
+  }
+}
+
+.page__body {
+  margin: 0;
+}
+
+.page__header {
+  z-index: 0;
```
alexander-ignatow, line 59: nesting :)

alexander-ignatow (changes requested):
Good job 👍 

Lets improve it even further by fixing issues with file structure

Please use nesting in scss where possible - it makes code way more readable. Update all files with and without the comments - I have not marked all places for the improve

Also, you have some issues with blocks. BEM blocks are meant to be reused, so you should really avoid setting the external styling on them. In practice that means mostly avoiding margins, height, and width (but there are some reasonable exceptions). Please see my comments and fix this too.

src/styles/blocks/menu.scss
PiekharievaK: Sometimes, out of habit, I write each class separately. I forgot to change it to nesting before making the pull request. Thanks for your comments.

alexander-ignatow (approved):
Good job! 👍

PR #2054, opened 2025-02-08
create landing page

OksanaKuziv:
[DEMO LINK](https://OksanaKuziv.github.io/layout_landing-page/).


KosolapovAnatolii (changes requested):
Your code looks very nice! 🆒 😎 
Just fix issues, that I show in this video: <https://www.loom.com/share/27858cfecb59427cbabe390cb482c5a8?sid=56830754-0440-45dc-82c9-48f2ea9ed31c>

src/styles/layout/about.scss
```diff
@@ -0,0 +1,61 @@
+.about {
+  display: grid;
+  row-gap: 48px;
+  background-color: $color-background;
+  margin: 0 -20px;
```
alexander-ignatow, line 5: Avoid margins on BEM blocks

src/styles/layout/contact.scss
```diff
@@ -0,0 +1,116 @@
+.contact {
+  @include content-padding-inline;
+
+  margin-bottom: 64px;
```
alexander-ignatow, line 4: Add this margin on the element level if needed
BEM blocks should avoid having "external" styling

src/styles/layout/contact.scss
```diff
@@ -0,0 +1,116 @@
...
+      height: 48px;
+      border-radius: 8px;
+      font-size: $font-size-sm;
+      background-color: $color-background-input;
+      border: 1px solid transparent;
+      transition: all 0.3s ease;
+      padding: 16px;
+      box-sizing: border-box;
+      outline: none;
+
```
alexander-ignatow, line 48: It is acceptable in your case, but a better idea would be styling via classes, not tag names

And in case you have many inputs in your app, take this styling to a separate BEM blocks

But it is OK in your app - no need to fix this now

src/styles/layout/section-title.scss
```diff
@@ -0,0 +1,22 @@
+.section-title {
+  margin: 0;
+  margin-bottom: 48px;
```
alexander-ignatow, line 3: Avoid margins on BEM blocks

alexander-ignatow (changes requested):
Good progress!

Still, some more issues are there

image

Hint: you can add something like 
```
input {
  font-family: ...;
}
```
to main.scss, but be careful, as it will affect all inputs

PR #2068, opened 2025-02-11
Layout landing page

ByatesEduard:


anastasiiavorobiova (changes requested):
Consider deploying your solution and providing a link in the PR description

PR #2071, opened 2025-02-11
create landing page bike

Ezero1:
[DEMO LINK](https://Ezero1.github.io/layout_landing-page/).


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
image

2. The header is too big, fix the size for background image 
image

3. You need to show additional phone number when hovering this icon and make a call when user clicks here 
image

4. Remove this padding 
image

5. Add hover effects for all images 
image

image

volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. Use the background image size for header, check it on the design 
image

2. Where is the horizontal indents here? 
image

3. Check the background image position here 
image

etojeDenys (approved):
lgtm

PR #2076, opened 2025-02-13
code landing-page

entodasha:


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. The background image should stretch to the full width of the browser window
image
image

2. Need to remove scroll when the menu is open 
image

3. The user cannot be able to change size of the textarea field 
image

4. Change the autofill styles for the form fields 
image

5. After submit the form don't need to refresh page and scroll to top, you need only reset form fields

entodasha:
> 5. After submit the form don't need to refresh page and scroll to top, you need only reset form fields

How can i do it without js?

entodasha:
> 1. The background image should stretch to the full width of the browser window
image
 
when i remote max width, header bg looks bad. this image is simply not meant foe screens larger than 1260px. but footer realy looks better.

> 2. Need to remove scroll when the menu is open

For solve this problem, i use z-index, but it does not remove scroll. Just make is no visible. Mayde you can recommend to another way for solve it.

VitaliyBondarenko1982 (changes requested):
Header and footer background images should be on full screen width.
If you have questions feel free to ask in chat.

etojeDenys (approved):

PR #2081, opened 2025-02-15
Develop

Kovtun-Dmytro:


Kovtun-Dmytro:
- Deploy and create a Pull Request with a [DEMO LINK](https://Kovtun-Dmytro.github.io/layout_landing-page/).

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
image

2. Don't need to block scroll after clicking on the menu links 
image

3. The font family should be are same everywhere 
image

4. Change the autofill styles for the form fields, the default and autofill styles should be are same 
image

5. After submit the form don't need refresh page and scroll to top, you need only reset form fields 
image

6. You need to make a call when user clicking here 
image

7. Open this additional link in the new tab 
image

8. Add hover effects for all clickable elements 
image

volodymyr-soltys97 (changes requested):
Almost done! 
To improve:
image

2. You need to separate these words with space `My Bike` 
image

3. Need to fix background image position for tablet, check it on the design 
image

volodymyr-soltys97 (approved):
Good job 👍

PR #2086, opened 2025-02-17
landing page

VentoDor:
[DEMO LINK](https://VentoDor.github.io/layout_landing-page/)
[B&O](https://www.figma.com/file/DtkQmQ797hk0nI4KfMi2Uq/BOSE-New-Version?type=design&node-id=6817-212&t=ZTV6Gl8NzaWkJ4FK-0)


VentoDor:
Good afternoon! Sorry for the inconvenience, I fixed the error in the settings, the demo link is working now





> Hi, your demo link isn't working, you need to make a deploy and check after the demo image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. You need to show the additional phone number when hovering this icon 
image

2. Add some hover effects for images 
image

3. Fix the default background color for this button 
image

4. The font family should be are same 
image

5. The user cannot be able to change size of these fields 
image

VentoDor:
I fixed all the points, but in point 4 the fonts seem to be the same, as I have a mixin for all fields setting the font as well 

> Good job 👍 To improve:
> 
> 1. You need to show the additional phone number when hovering this icon
> 
> image
> 2. Add some hover effects for images
> 
> image
> 3. Fix the default background color for this button
> 
> image
> 4. The font family should be are same
> 
> image
> 5. The user cannot be able to change size of these fields
> 
> image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. The phone number position and styles looks different as on the design, need to fix it 
image

image

3. Not fixed 
image

4. Add cursor pointer for this button 
image

image

VentoDor:
image
image

I'm sorry, but I don't fully understand what I need to do in the 3rd point, I thought I just need to remove the extra style when hovering and set the color of the button.

In the second point I decided that then it is better not to enlarge, but to add a shadow, this rule is not in the design anyway.

volodymyr-soltys97 (changes requested):
Hi, you need to fix button as on the design 
image

anastasiiavorobiova (approved):

PR #2099, opened 2025-02-24
add task solution

vmohnach:
[DEMO LINK](https://vmohnach.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
image

2. You need to show additional phone number when hovering this icon and make a call when user clicks here 
image

image

image

5. The font family should be are same everywhere 
image

6. Make these fields are required before sending 
image

image

vmohnach:
Need to disable scroll when the menu is open
В мене не виходить це зробити, спробував добавити overflow: hidden; до menu, потім до page__menu і нічого не спрацювало

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add the topical title on the page 
image

2. Open this additional link in the new tab 
image

3. Add hover effects here 
image

image

etojeDenys (approved):
lgtm

PR #2104, opened 2025-02-26
Landing Page

ivsasha:
[DEMO LINK](https://ivsasha.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
image

2. Fix hover effects for this icon 
image

3. The background image is currently very heavy, which is why it takes a long time to load, try optimizing it a bit
image

4. The left indent should be are same everywhere 
image

image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. The additional block looks different, need to check it on the design 
image

image

image

4. The gap between these blocks should be are same 
image

5. These elements should be links 
image

VitaliyBondarenko1982 (approved):
Looks good

PR #2107, opened 2025-02-27
Develop

tarasholotiuk:
https://tarasholotiuk.github.io/layout_landing-page/


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
image

2. Need to show additional phone number when hovering this icon 
image

image

4. Add hover effects to images 
image

5. Add cursor pointer to this button 
image

6. These elements should be links and add hover effects here 
image

vadiimvooo (changes requested):
Good job, there are a few changes to do. Please feel free to ask for help in fe_chat.
1. Add scroll if elements height is greater then viewport height
image
2. Element on hover should not overlap neighboor elements
image

vadiimvooo (approved):
Well done

PR #2111, opened 2025-03-02
layout_landing-page B&O project

Dmytro-Nyzhnyk:
[DEMO LINK](https://Dmytro-Nyzhnyk.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Hi, your demo link isn't working, you need to make a deploy and check after the demo link 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
image

image

3. Add transition when the menu appears on the page and remove scroll when menu is open 
image

4. Add hover effects for all clickable elements on the page and for images
image
image

image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Change title on the page to `B&O`
image

2. Fix styles and position for the phone number, check it on the design 
image

3. These elements should be links and add hover effects here 
image

image

anastasiiavorobiova (approved):
Great work!

PR #2122, opened 2025-03-03
Develop

akyval:
- [DEMO LINK](https://akyval.github.io/layout_landing-page/)


akyval:
If u have something to tell me, please write me in private message Nazar Butrei(Назар Бутрей)

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add a favicon to the page 
image

2. Need to show additional phone numbers when hovering this icon, check it on the design 
image

image

4. Fix gap here 
image

5. Add transition for all hover effects on the page 
image

6. The horizontal indents should be are same everywhere 
image

7. Looks different as on the design, need to check and fix styles here 
image

anastasiiavorobiova (changes requested):
Great work!
- Screenshot 2025-03-05 at 13 14 41
- Screenshot 2025-03-05 at 13 14 47
- Check the layout for mobiles
- Screenshot 2025-03-05 at 13 13 59
- https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/
- Autocomplete background should be changed
- Screenshot 2025-03-05 at 13 12 45
- All hovers should be smooth

maxim2310 (changes requested):

transition works only on enter cursor, but not on leave 
image

VitaliyBondarenko1982 (approved):
.

PR #2134, opened 2025-03-05
landing page final version

liudmylapanchenko:
[DEMO LINK](https://liudmylapanchenko.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page:
1. Need to show additional phone number when hovering this icon 
image

2. The user cannot be able to change size of the textarea field 
image

image

image

maxim2310 (changes requested):

check font of it 
image

anastasiiavorobiova (approved):
Awesome work!

PR #2137, opened 2025-03-05
Develop

vikysichkaa17:
Preview/Demo
https://vikysichkaa17.github.io/layout_landing-page/


maxim2310 (changes requested):

1. it looks like icons are too big, check plz according design 
image
2. block shouldn't overlap nearest ones 
image
3. there is not enough space 
image
4. remove default styles here 
image

maxim2310 (changes requested):

1. better add cursor pointer on interactive elements 
image
2.user should be able to click on this buttons 
image

vikysichkaa17:
Thank you for your advice! Could you check it again?

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Need to hide additional block with phone numbers on the mobile and tablet versions and make a call when user clicking on the phone icon 
image

2. The mobile version looks strange, need to fix it 
image

3. Add hover effects here 
image

image

vikysichkaa17:
Think it works ok now.  Thank's!

volodymyr-soltys97 (approved):
Great! 🔥

PR #2143, opened 2025-03-07
Solution

xxxKoRoBoKxxx:
[DEMO LINK] - https://xxxkorobokxxx.github.io/layout_landing-page/


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Fix styles and position for this phone number, check it on the design 
image

2. The prices should be are on the same horizontal line 
image

3. Not fixed from the previous review 
image

xxxKoRoBoKxxx:
Hi
Some notes:
1. Images worked properly on all my devices after first fix
image
image
2. I decided to make background for phone number because we can baraly see black text with hair on back of it
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. The phone number styles looks different as on the design, need to check it on the design and to fix 
image

2. Not fixed from the previous review 
image

etojeDenys (approved):
lgtm

PR #2144, opened 2025-03-07
Develop

Darraya:
[DEMO LINK](https://Darraya.github.io/layout_landing-page/).


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Need to make a call when user clicking here 
image

2. Need to fix placeholder text position 
image

3. Fix this error after submit the form 
image

image

5. Add transition when this menu appears on the page 
image

PR #2152, opened 2025-03-11
The Met

Alya-ya:
[DEMO LINK](https://Alya-ya.github.io/layout_landing-page/).


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add hover effects for all clickable elements 
image

2. Add cursor pointer here 
image

image

image

5. Need to make a call when user clicking here
image
 
 6. Check gap between these icons on the mobile and tablet versions 
image

volodymyr-soltys97 (approved):
Great! 🔥

PR #2159, opened 2025-03-16
Develop

ovaheman:
test


volodymyr-soltys97 (changes requested):
Hi, there is no demo link. Also you need to provide demo link to the PR description 
image

ovaheman:
[DEMO LINK](https://ovaheman.github.io/layout_landing-page/).

maxim2310 (changes requested):

1. check background img size
2. here should be clickable number on hover 
image
3. check font 
image
4. remove scroll when menu is opened
5. need to align it 
image
6. would be nice to add hover effect on it 
image
7. check width and indent here 
image
8. these also should be links 
image
9. wrong font and indent 
image
10. inputs in form should be required

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. This element should be link with `mailto` 
image

image

3. Check the background image position on the mobile and tablet versions 
image
image

VitaliyBondarenko1982 (approved):
Looks good to me.

PR #2160, opened 2025-03-16
add task solution

DSOst:
-[DEMO LINK](https://DSOst.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Need to remove scroll when the menu is open 
image

2. Add hover effects to images 
image

3. Check gap between these elements as on the design 
image

4. The user cannot be able to change size of the textarea field 
image

5. Fix this error after submit the form 
image

image

7. These elements should be links 
image

8. Add hover effects here 
image

9. Hide the additional phone number on the mobile and tablet versions 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Need to disable scroll when the menu is open 
image

2. The images shouldn't overlap each other when hovering. Also, the image should scale down as smoothly as it scales up. 
image

3. Still not fixed 
image

4. Add cursor pointer for this button 
image

5. Add the correct address for these links 
image

maxim2310 (changes requested):

1. not fixed 
image
2. not fixed 
image
3. 
image
4. fix hover effect 
image

vadiimvooo (changes requested):
Good job, there are few changes to do. Plase feel free to ask for help in fe_chat.
1. Fonts should be the same
image
2. On form submit you should not update page
3. Add transition on all interactive elements. Transition should be also when element is getting back to it's original state as well.

PR #2162, opened 2025-03-18
Develop

vvyboichyk:
1. In this project, a website was created for the Bang & Olufsen company, which specializes in the development and production of audio and video systems.
2. In this project, the following technologies were used: Flex Box, BEM, SCSS, Mixins, Functions, Transformations and Animations, Grid.
3. https://vvyboichyk.github.io/layout_landing-page/
4. - [B&O](https://www.figma.com/file/DtkQmQ797hk0nI4KfMi2Uq/BOSE-New-Version?type=design&node-id=6817-212&t=ZTV6Gl8NzaWkJ4FK-0)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Need to remove scroll when the menu is open 
image

2. Add hover effects to images 
image

image

4. These elements should be links and add hover effects here 
image

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. Need to make a call when user clicking here or phone number
image
image

PR #2165, opened 2025-03-19
Develop

bogdank0:
[DEMO LINK](https://bogdank0.github.io/layout_landing-page/).


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add a favicon to the page 
image

image

image

image

5. Change the autofill styles for the form fields. The autofill and default styles should be are same 
image

6. Hide this additional block with phone numbers on the mobile and tablet versions. Also need to make a call when user clicking on the phone icon 
image

vadiimvooo (changes requested):
Good job, there are a few changes to fix. Please feel free to ask for help in fe_chat.
1. Hovered element should not overlap neighbor elements.
image
2, On form send you should not reload page.

volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. This comment still not fixed from the previous review. Also you need to reset form fields after clicking send
image

2. Hide the additional block with phone numbers on the mobile and tablet versions 
image

3. Make a call when user clicking on this icon 
image

volodymyr-soltys97 (approved):
Great! 🔥

PR #2167, opened 2025-03-21
Add task solution

s1rserg:
[DEMO LINK](https://s1rserg.github.io/layout_landing-page/)
[MOCKUP](https://www.figma.com/design/DtkQmQ797hk0nI4KfMi2Uq/BOSE-New-Version?node-id=6802-139&p=f&t=mJ0WfOqkaP1vGiMn-0)


VitaliyBondarenko1982 (changes requested):
Good job.
Suggest a couple improvements.
- prevent resize for textarea
Screenshot 2025-03-22 at 09 29 42
- add cursor pointer for send button. Check other interactive elements.

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. The font styles looks different as on the design, need to fix it. It should also be possible to hover over a phone number
image

2. The prices should be are on the same horizontal line 
image

image

s1rserg:
> 1. The font styles looks different as on the design, need to fix it. 

I changed the font so it is as on the design, but I hope you are not suggesting that I should get rid of background-color on it. The phone number on wider screens is unreadable without it due to header background image.

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. This comment still not fixed from the previous review. Check styles on the demo link 
image

PR #2172, opened 2025-03-25
add task solution

DmutroSheva:
[DEMO LINK](https://DmutroSheva.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
image

image

3. Need to open the menu when user clicking here 
image

4. The gap between these elements looks different as on the design, also check font styles here 
image

5. The horizontal indents should be are same everywhere 

6. Add hover effects for all clickable elements on the page

PR #2190, opened 2025-03-30
add task solution

OleksandrLib:
-[DEMO LINK](https://OleksandrLib.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. show number on hover
image

2. different colors here
image

3. images in low res, try to re-import them
image

4. you need to stretch it to full width
image

5.  looks broken
image

6. should be textarea tag
image

PR #2194, opened 2025-03-30
add solution

ATkachenkoA:
[DEMO LINK](https://ATkachenkoA.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. fix the color mismatch here
image

2. number should appear on hover
image


3. color should be yellow
image

4. form looks different in figma
image

VitaliyBondarenko1982 (changes requested):
Good job.
Almost done.
On any width of screen exists horizontal scroll.
Reason in 100vw - try do not use it.
100vw it is width of screen which not count width of scrollbar
Screenshot 2025-04-12 at 10 49 51

maxim2310 (changes requested):

1. icon and logo should be in svg 
image
2. background for img and font style 
image
3. also background 
image
4. font wrong here 
image
5. backgrounds 
image
6. submit button should not reload page just reset form

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. The images shouldn't overlap each other when hovering 
image

2. Images should scale down as smoothly as they scale up on hover.
image
image

3. Make these fields required before sending 
image

4. This element should be link with `mailto` 
image

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. Remove hover effects from this block 
image

2. The color should be same everywhere 
image

3. After submit the form don't need refresh page and scroll to top, you need only to reset form fields 
image

etojeDenys (approved):
good job

PR #2200, opened 2025-03-31
add task solution

IvanKovalyoff:
[DEMO LINK](https://IvanKovalyoff.github.io/layout_landing-page/).


etojeDenys (changes requested):
1. disable the scroll if the menu is opened
image

2. disable the textarea resizing
image

3. add the `required` attribute to all fields in the form
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. The gap and font styles looks different as on the design, need to fix it 
image

2. Check width for the form fields as on the design 
image

3. Hide the additional phone number on the mobile and tablet versions 
image

4. The font styles looks different 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Need to fix position for phone number, check it on the design 
image

2. Font styles looks different here as on the design 
image
image
image
image
image

3. The images shouldn't overlap other elements when hovering 
image

image

5. Change the autofill styles for the form fields. The autofill and default styles should be are same. Also, remove a default blue border
image

etojeDenys (changes requested):
image
fix the horizontal scroll on mobile

volodymyr-soltys97 (approved):
Great! 🔥

PR #2202, opened 2025-04-01
add task solution

AndrewPrszn:
[DEMO LINK](https://AndrewPrszn.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. add the favicon
image


2. show number on hover
image


3. stretch these sections on full width
image


4. looks different from mockup
image

etojeDenys (changes requested):
form should match the styles from the mockup
image


2. limit the width in the block to match the rest of the site
image

3. looks weird
image

4. fix the horizontal scroll
image

maxim2310 (changes requested):

1. add some space here, check dont and add posibility to click on it and initiate call 
image
2. check font here 
image
3. imgs shouldn't overlap each other
image
4. check line height here 
image
5. when i hover on some input others are jumping 
image
6. page should reset form, but not reload page 
image
7. check format according design and it should lead us to google map 
image

VitaliyBondarenko1982 (approved):
Looks good.

PR #2204, opened 2025-04-02
add task solution

Anatoli37:
[DEMO LINK](https://Anatoli37.github.io/layout_landing-page/).


maxim2310 (changes requested):


image
2. add some space here and check the font 
image
image
image

3. you need to fix it on mobile  
image
4. page shouldn't reload after submit, submit button should just reload the form 
5. remove default autofill styles 
image

anastasiiavorobiova (changes requested):
Great work! Check fonts here 
Screenshot 2025-04-16 at 14 11 32

VitaliyBondarenko1982 (changes requested):
Almost done.

[checklist point 15](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md)

etojeDenys (approved):

PR #2246, opened 2025-04-18
First Solution

DimaK123:
https://dimak123.github.io/layout_landing-page/


etojeDenys (approved):
good job, but it's better to disable the scroll if the menu is opened

PR #2250, opened 2025-04-18
initial commit

yelyzavetaz:
[DEMO LINK] - https://yelyzavetaz.github.io/layout_landing-page/


BudnikOleksii (changes requested):
Good work 👍 
A few moments should be fixed in accordance with checklist and design:
- Use the logo as a favicon
image
- Make sure everything looks neat on mobile and without horizontal scrolling. On mobile some content has different size and some sections are not centered
image

image
- Also, on desktop some sections have the same problems
image
image

ArturErmolenko (changes requested):
Almost done 🔥 
Let's work with form(
we want the form to submit but now the submit button just scrolls me to the header):

1. All form fields should be required and have correct types (email, tel, etc.)
2. The page shouldn't be reloaded on form submit, but the fields should be cleared.
Look at [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md) 13 -15

BudnikOleksii (approved):
Good work 👍

PR #2258, opened 2025-04-19
Nothing landing result

Stariivitalii:
https://Stariivitalii.github.io/layout_landing-page/


BudnikOleksii (changes requested):
Good work 👍 
Just fix buttons a bit:
- This button on desktop is too big, check the design
image
- Button color change isn't smooth, set transition duration
- All interactive elements such as button should have a cursor pointer

BudnikOleksii (approved):
Good work 👍

PR #2261, opened 2025-04-20
Develop

Nnazariypanochko:
[DEMO LINK](https://nnazariypanochko.github.io/layout_landing-page/).


etojeDenys (changes requested):
1. stretch menu to the full width
image

2. change the title and add the favicon
image

3. number should appear on hover
image

4. don't forget to follow all points from this file https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. The phone number position looks different as on the design, need to fix it 
image

2. Need to disable scroll when the menu is open and add transitions when this menu appears on the page 
image

image

image

5. Make these fields required before sending 
image

etojeDenys (approved):
looks good, but don't forget to disable the scroll if the menu is opened

PR #2262, opened 2025-04-21
Develop

YarRudenko:
[DEMO LINK](https://YarRudenko.github.io/layout_landing-page/).


src/styles/blocks/contacts.scss
```diff
@@ -0,0 +1,136 @@
...
+
+     @include on-desktop {
+
+      grid-column: 8 / span 5;
+    }
+  }
+
+  &__input {
+    color: #7E7E83;
+    background-color: #F9F9F9;
```
ArturErmolenko, line 45: variable

ArturErmolenko (changes requested):
Code looks good 👍 
Just use more variables for repeating colors etc

I can't check your work because smth bad happened with Demo Link, let's fix it
Screenshot 2025-04-24 at 13 33 25

BudnikOleksii (changes requested):
Good work 👍 
- When user hovers a phone icon you should show a phone number
image
image
image
image

BudnikOleksii (approved):
Nicely done 👍

PR #2275, opened 2025-04-22
first code

yuliimar:
[DEMO LINK](https://yuliimar.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add a favicon to the page 
image

image

image

4. Fix gap between these elements 
image

image

6. External link should be open in the new tab 
image

7. Add hover effect to clickable elements 
image

8. The header looks different as on the design, need to fix 
image

volodymyr-soltys97 (changes requested):
Hi, the comments from the previous review still not fixed, you need to make a deploy again and check after the demo

etojeDenys (approved):
good job

PR #2279, opened 2025-04-22
final code

nazukvv:
[DEMO LINK](https://nazukvv.github.io/layout_landing-page/).


maxim2310 (changes requested):

1. here should be clickable number on hover 
image
2. prices should be aligned 
image
3. would be nice to have some hover effect on these links 
image
4. font is different 
image
5. page should not be reloaded after clicking on submit button. but just reset form

etojeDenys (approved):
👍

PR #2280, opened 2025-04-22
add task solution

aholubko:
[DEMO LINK](https://aholubko.github.io/layout_landing-page/)


maxim2310 (changes requested):
Demo looks broken, check plz 
image

maxim2310 (changes requested):

1. click on this should initiate call and be visible Only on hover on icon call, not on epace
image
2. looks like icons are bigger then on design, check plz 
image
3. prices should be aligned 
image
4. images shouldn't overlap each others 
image
5. remove default styles 
image

maxim2310 (changes requested):
something went wrong 
image

anastasiiavorobiova (approved):
Awesome work!

PR #2284, opened 2025-04-23
Develop

Ivan7419:
[DEMO LINK] (https://Ivan7419.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add hover effects to images 
image

2. All these elements should be links and add hover effects here 
image

image

4. The user cannot be able to change size of these fields 
image

etojeDenys (approved):
all inputs should have `required` attribute
image

PR #2292, opened 2025-04-25
Develop

LiSenber:
[DEMO LINK] (https://LiSenber.github.io/layout_landing-page/)


maxim2310 (changes requested):

1. here should be clickable number 
image
2. add title and favicon 
image
3. prices should be aligned and with bold font 
image
4. would be nice to add hover effect
image
 5. remove scroll while menu is opened 
image
6. add smooth scroll on click on anchor links 
7. fix width according to design 
image
8. remove default style 
image

VitaliyBondarenko1982 (changes requested):
Good job.
Almost done.
- Ensure that all navigation links work smoothly (now after click on some menu item no smooth scrolling on page)

etojeDenys (changes requested):
your demo link doesn't work, try to fix that and then re-request the review

etojeDenys (approved):

PR #2296, opened 2025-04-25
Solution

panasovskyi:
https://panasovskyi.github.io/layout_landing-page/


VitaliyBondarenko1982 (changes requested):
Good job.
Needs some improvements.

- prevent resize text area
Screenshot 2025-04-26 at 10 20 29

- click on send button leads on 405 page. The page shouldn't be reloaded on form submit, but the fields should be cleared. Use:

VitaliyBondarenko1982 (approved):
Looks good.

PR #2300, opened 2025-04-28
THE MET MUSEUM

Kaminazer:
Finish task layout_landing-page.

An adaptive MET Museum page implemented according to [figma design](https://www.figma.com/design/lSR1m42L9YwzQwzzxKwHpw/THE-MET?node-id=8590-160&t=0DEeXfdDErhnVoOc-0)

To see a demo click [here](https://kaminazer.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. The user cannot be able to change size of the textarea field 
image

2. Change the autofill styles for the form fields. The autofill and default styles should be are same 
image

3. You need to make a call when user clicking on the phone icon 
image

VitaliyBondarenko1982 (changes requested):
Good job.
Almost done.
Check styles for header background.
Exists some horizontal lines and looks like background image is repeating.
It should have backgoround-repeat: no-repeat.

volodymyr-soltys97 (approved):
Good job 👍

PR #2303, opened 2025-04-28
Develop

Sofiia13:
[DEMO LINK](https://Sofiia13.github.io/layout_landing-page/)


maxim2310 (changes requested):

1. user should be able to click on it 
image
2. remove scroll when menu is opened 
image
3. would be nice to add hover effect 
image
4. add transition here 
image
5. submit button should reset form 
image
6. these should be links 
image

maxim2310 (changes requested):

1. icons and logo should be in svg
image
2. all effects and animations should be with transition
image
3. change color also with transition 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Make these fields required before sending 
image

image

image

etojeDenys (changes requested):
1. should be type=submit
image

maxim2310 (changes requested):
submit button should reset form and not reload page

volodymyr-soltys97 (approved):
Good job 👍

PR #2311, opened 2025-04-30
add task solution

Whslv:
[DEMO LINK](https://Whslv.github.io/layout_landing-page/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Need to show additional phone number when hovering this icon, check it on the design 
image

2. This text looks different as on the design 
image

image

4. The placeholder text should be same as on the design 
image

5. Make these fields required before sending 
image

6. The user cannot be able to change size of the textarea 
image

7. This element should be link with `mailto` 
image

etojeDenys (approved):
don't forget to add this to your form
image

2. good job overall

PR #2327, opened 2025-05-06
layout_landing-page

MarinaPrague:
- Use the logo as a favicon
- Make sure everything looks neat on mobile and without horizontal scrolling
- The speed of animations should be the same throughout the page (for example, increasing when hovering or moving blocks when scrolling)
}
.menu {
}
.menu__top {
  top: 0;
  z-index: 1;
  background-color: #e5e5e5;
}
- Placeholders should give examples of what should be added, to make the expected format clear

### DIFF
--- a/src/styles/blocks/icon.scss
+++ b/src/styles/blocks/icon.scss
@@ -18,23 +18,19 @@
     background-image: url(../images/icons/icon-phone-2x.png);
 
     &:hover {
-      cursor: none;
       width: 179px;
       background-image: url(../images/icons/icon-phone-number-2x.png);
     }
-
-    @include hover-cursor('../images/icons/cursor-mouse.png');
   }
 
   &--menu {
     background-image: url(../images/icons/icon-burger-menu-2x.png);
 
     &:hover {
-      cursor: none;
       background-image: url(../images/icons/icon-burger-menu-hover-2x.png);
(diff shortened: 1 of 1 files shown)


etojeDenys (changes requested):
don't forget to provide the demo link and then re-request the review

maxim2310 (changes requested):
instead of this, here should be your demo link like in tasks before 
image

MarinaPrague:
https://MarinaPrague.github.io/layout_landing-page/

etojeDenys (changes requested):
1. do not display the cursor on icon on hover
image

2. display the number on hover
image

3. stretch this block on full width
image

4. compare the gaps with gaps in figma
image


5. disable the textarea resizing
image

6. center or stretch the width of text
image


7. same with button
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Remove additional cursor here when hovering 
image
image

2. Add a better favicon to the page 
image

3. Fix font weight for the prices, check it on the design 
image

4. The textarea height looks different as in the design 
image

5. Add cursor pointer for all clickable elements 
image

volodymyr-soltys97 (approved):
GJ! 👍

PR #2346, opened 2025-05-11
all done V 2.0

Bo-oV:


etojeDenys (changes requested):
don't forget to provide the demo link and then re-request the review

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Need to show additional phone number when hovering this icon 
image

image

3. Fix font weight for the prices 
image

4. Add hover effects to images 
image

5. Looks different as in the design 
image

image

7. Add cursor pointer for all clickable elements 
image

8. These elements should be links and add hover effects here 
image

maxim2310 (changes requested):

1. all icons and logo should be in svg 
image
2. not fixed 
image
3. cards should be aligned 
image
4. animation is jumping when i leave mouse from item 
image
image
5. add shadow on whole card and add some border radius 
image
6. this should be links
image
7. not fixed 
image
image
8. form should not reload page after submit

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. This comment still not fixed, check it on the design 
image

2. Not fixed 
image
image

3. Add transition for hover effects everywhere 
image

4. Fix the button width 
image

image

Design 
image

etojeDenys (approved):
good job

PR #2353, opened 2025-05-12
Develop

dNaiko:
[DEMO LINK](https://dNaiko.github.io/layout_landing-page/)


maxim2310 (changes requested):

1. main img looks cut
image
2. icons and logo should be in svg 
image
3. remove scroll when menu is opened 
image
4. here should be number on which you should be able to click
image
5. add some hover effect here 
image
6. when i hovering on inputs all form is jumping, check it plz 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add a favicon to the page 
image

2. The additional phone number looks different as on the design, check it on the design 
image

3.

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. Need to hide this additional block with phone numbers on the mobile and tablet versions 
image

2. Fix the gap between these block and phone icon, check it on the design 
image

etojeDenys (approved):
lgtm
but align these icons
image

PR #2354, opened 2025-05-12
Solution

NattanJohn:
[DEMO LINK](https://NattanJohn.github.io/layout_landing-page/)


pedro-ruas (changes requested):
Just one thing missing from the checklist:
- Text beside the Form should be links that send the user to telephone, email and maps with address

pedro-ruas (approved):

PR #2356, opened 2025-05-13
add task solution

WanderleiBolzan:
Deploy and create a Pull Request with a [DEMO LINK](https://WanderleiBolzan.github.io/layout_landing-page/)


pedro-ruas (changes requested):
Page differs from the design:
- Images are not correctly positioned,
- Menu, Form and other elements are positioned as like Mobile view when in Desktop view,
- Missing correct implementation of the form submit and required fields,
- Some links do not work

PR #2368, opened 2025-05-20
add task solution

leo15782:
- [DEMO LINK](https://leo15782.github.io/layout_landing-page/)


DanielCamposDevX (changes requested):
The page is looking good but it has some points of imporvemente.
Folowing the layouts passed on task description is good, although if you use the max width on the project, screens in full hd for example are gonna appear as broken, so try to not limit your width and remember to center elements(maybe using a mix of flex and grid on your page)

pedro-ruas (changes requested):
Ok, looks good and works mostly.
Two checklist points missing:
- The phone number should appear when hovering the phone icon at page top
- The contact form should not allow to be sent if every fields isn't complete

DanielCamposDevX (approved):

PR #2369, opened 2025-05-20
primeira versao

LuisCampelo2:
[DEMO LINK] https://LuisCampelo2.github.io/layout_landing-page/


DanielCamposDevX (changes requested):
Page looks perfect on mobile, but theres is some points about the checklist itself.
On your inputs placeholder, you are not using real examples like instead of "email" use "email@email.com".
There is no hover effects on the page (Images header & menu should have some hover effects like said on the checklist of this task.

PR #2370, opened 2025-05-20
Develop

LuisCampelo2:
[DEMO LINK] https://LuisCampelo2.github.io/layout_landing-page/


pedro-ruas (changes requested):
Luis, I don't see errors in your implementation, it is well done.
However, if you read carefully the checklist and task description, it says **explicitly not to do the Bose page**, and instead, pick one of the other four possibilities.
This is why I'm not approving this task

PR #2390, opened 2025-05-27
Page

LizaN-I:
https://lizan-i.github.io/layout_landing-page/src/


etojeDenys (changes requested):
make sure you provide a demo link and request a review again

maxim2310 (changes requested):

1. icons and logo should be in svg 
image
2. here should be phone number on hover 
image
3. align it 
image
4. add hover, cursor pointer and some action on button 
image
5. these should be links 
image
6. textarea should have fixed sizes 
image

PR #2399, opened 2025-05-31
add task solution

sidond:
- [DEMO LINK](https://sidond.github.io/layout_landing-page/)


etojeDenys (changes requested):
disable textarea resizing
image

2. should be links
image

3. add the `required` attribute to all fields
image

4. it's better to export icons as .svg
image

PR #2402, opened 2025-06-02
Develop

isapq:
- When done check yourself using the [CHECKLIST](./checklist.md) when finished.
- Deploy and create a Pull Request with a [DEMO LINK](https://isapq.github.io/layout_landing-page/).


DanielCamposDevX (changes requested):
All images should have hover.
Phone Icon should show the phone number on hover.
When submitting the form should clean the fields.
Remember to change the cursor no clickable buttons.
Adresses should be links to google maps
 phone numbers should start a call on click

src/styles/packsOsStyles/sectionTheDetails.scss
```diff
@@ -0,0 +1,286 @@
...
+}
+
+/* application of hover on images */
+
+img {
+  transition: 0.3s;
+}
+
+.sectionMain__details-container-img1:hover,
+.sectionMain__details-container-img2:hover {
```
raulriato, line 42: Your classes name are a little bit confusing. You are applying this transform on these two classes, but it's not applying to a lot of other images on your pages. Since all the images from the same section are equal, you don't need a separate class for each of them, just one class that is applyed for all of them.

src/formClear.js
```diff
@@ -0,0 +1,26 @@
...
+  if (
+    campName.value !== '' ||
+    campEmail.value !== '' ||
+    campMessage.value !== ''
+  ) {
+    campName.value = '';
+    campEmail.value = '';
+    campMessage.value = '';
+  }
+});
```
raulriato, line 26: you don't need an eventListener to deal with this. You just need to add an attribute `onsubmit` to the `<form>` tag.

src/index.html
```diff
@@ -15,8 +28,387 @@
...
+      </section>
+
+      <section
+        class="sectionMain__contactUs"
+        id="contact"
+      >
+        <h2 class="sectionMain__contactUs-titleSection">Contact US</h2>
+
+        <div class="sectionMain__contactUs-contactUs">
+          <form class="sectionMain__contactUs-contactUs-data">
```
raulriato, line 340: add the attribute `onsubmit` here, instead of using an eventListener. The listener isn't even working. If you want to test it, type everything and click on `Send`, it will reload the page.

src/index.html
```diff
@@ -15,8 +28,391 @@
...
+      <section
+        class="sectionMain__contactUs"
+        id="contact"
+      >
+        <h2 class="sectionMain__contactUs-titleSection">Contact US</h2>
+
+        <div class="sectionMain__contactUs-contactUs">
+          <form
+            class="sectionMain__contactUs-contactUs-data"
+            onsubmit="return forceTofillinEverything();"
```
pedro-ruas, line 342: You should also prevent the page reload. As it's required in the checklist.
For this, you can use something like:
```
onsubmit="(event) => {
  event.preventDefault();
  return forceToFillinEverything();
}"
```

pedro-ruas (changes requested):
Missing prevent page reload on form submit

PR #2404, opened 2025-06-03
Finished design

AV1788:
[DEMO LINK](https://<your_account>AV1788.github.io/layout_landing-page/).


etojeDenys (changes requested):
Almost done

1. add hover to logo

2. add a favicon and title to the page
image

3. photos should not overlap when hovering
image

4. it would be great to add a delay to the hover effect here
image

PR #2405, opened 2025-06-04
Second Commit

Juliovergueiro:


Juliovergueiro:
> Please provide a working DEMO link so we can test this page. Also, why are there so many static fonts? You are not using them all

Sorry Pedro, here it is: https://Juliovergueiro.github.io/layout_landing-page/

Would you like me to do a new pull request without the unused fonts and include the DEMO link in Readme.md?

pedro-ruas (changes requested):
Julio, a few things missing:
- Images should have a transition, not only "jump" on hover, but with a animation duration (this is on the checklist)
- Your form should not be able to be sent when it is incomplete. All fields are required.
- Besides that, the page should NOT reload when sending the form
- Each contact, phone number, addresses, and emails, located beside the form, should be working links

Juliovergueiro:
> Julio, a few things missing:
> 
> * Images should have a transition, not only "jump" on hover, but with a animation duration (this is on the checklist)
> * Your form should not be able to be sent when it is incomplete. All fields are required.
> * Besides that, the page should NOT reload when sending the form
> * Each contact, phone number, addresses, and emails, located beside the form, should be working links

Thanks for the review, I think it's working now

pedro-ruas (approved):

DanielCamposDevX (approved):

PR #2419, opened 2025-06-13
Develop solution

Asyyaab:
Deploy and create a Pull Request with a [DEMO LINK](https://Asyyaab.github.io/layout_landing-page/).

# Tips & Hints
- Check `background-image: url()` to be relative to the `main.scss`. So should start with `../images`.
- If some points from the checklist are not relevant to your design, just skip it.


etojeDenys (changes requested):
Good job

1. the phone number should appear when you hover over it
image

2. add a hover effect and change the cursor on the links
image

volodymyr-soltys97 (changes requested):
Hi, the previous comments still not fixed, you need to make a deploy and check the demo link after
Also, the user cannot be able to change size of the textarea field 
image

PR #15, opened 2023-05-27
bike landing

Dmytro-Kulyk-cpu:
[DEMO LINK](https://<Dmytro-Kulyk-cpu>.github.io/layout_landing/)

PR #2423, opened 2025-06-15
add task solution

CodableBase:
- [DEMO LINK](https://CodableBase.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. remove horizontal scrolling
image

2. it looks different with the layout
image

figma:
image

3. the phone number should show up when you hover over the icon. Also make icon phone link
image

4. it looks different with layout
your:
image

figma:
image

5. add a hover effect to the links and decrease the gap slightly
image

Anton-Kuchmasov (changes requested):
Good job!

To improve:
1. Check twice font-size of phone number (appears on `phone-icon` hover) in Figma & in your implementation:
image

2. Would you mind to make all those images interactive?
On `hover` it should be like `transform: scale(1.04)` with smooth transition
image

3. Add `cursor: pointer` to each single interactive element on your page (e.g. buttons, links...)
4. Clicking on `Send` button should only wipe all added data - no side effects like page reloading or scrolling allowed. Check out #15 of [this ](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md) instruction

5. Disable page scrolling when `aside` menu is opened. Check out #11 of instruction that I provided above

Anton-Kuchmasov (changes requested):
don't forget to make a new Deploy after changes - now your Landing looks the same

Anton-Kuchmasov (approved):
Well done!

PR #2430, opened 2025-06-16
Develop

svyatoslav1313:
[DEMO LINK](https://svyatoslav1313.github.io/layout_landing-page/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -344,15 +344,18 @@
                   type="text"
                   class="contact-form__input"
                   placeholder="Name"
+                  required
                 />
                 <input
                   type="email"
                   class="contact-form__input"
                   placeholder="Name"
+                  required
                 />
                 <textarea
                   class="contact-form__textarea"
                   placeholder="Message"
+                  required
                 ></textarea>
                 <button class="contact-form__button">Send</button>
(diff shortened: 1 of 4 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. The background image needs to be optimized, it currently takes up too much space and is taking a long time to load.
image

image

3. The prices font styles looks different as on the design, need to fix it 
image

4. Add some hover effects to images 
image

image

image

etojeDenys (changes requested):
1. there should be links
image

2. don't forget to implement these steps
image

volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. The phone number position looks different as on the design, need to fix it 
image

2. Font size for the prices looks different as on the design 
image

image

image

5. Fix this error after clicking on the phone icon, you need to use `tel` in href 
image

6. The height looks different as on the design 
image

7. Add the correct address on maps here 
image

8. Add hover effects here 
image

etojeDenys (changes requested):
Almost done.

1. make these inputs required
image

PR #2431, opened 2025-06-16
add task solution

Artem1501:
[DEMO LINK](https://Artem1501.github.io/layout_landing-page/).


etojeDenys (changes requested):
almost done

1. still look different with the layout
image

2. this block should have a different background color
image

3. the whole page has this space on the right.
image
image

PR #2433, opened 2025-06-16
Main

Ruslan-oss-max0:
- [DEMO LINK](https://Ruslan-oss-max0.github.io/layout_landing-page/)


anastasiiavorobiova (changes requested):
Great work!
- Screenshot 2025-06-17 at 19 35 13
- Consider disabling the scroll under the menu
- Screenshot 2025-06-17 at 19 34 10
- A user shouldn't be able to resize the textarea
- Consider just clearing the form after the submit instead of page reloading
- Screenshot 2025-06-17 at 19 34 07
- Consider using links here (`mailto`, `tel`, `Google Maps`)
- Screenshot 2025-06-17 at 19 33 54
- Check fonts here

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add the better favicon and change title ti `B&O` 
image

2. The font weight for the prices looks different as on the design, need to fix 
image

image

4. The font family for the placeholder text should be are same 
image

5. Change the autofill styles for the form fields, the autofill and default styles should be are same 
image

6. Add hover effects here 
image

7. Check gap between these elements 
image

8. The height for this button should be 56px, check it on the design 
image

9. Hide the additional phone number on the mobile and tablet versions 
image

volodymyr-soltys97 (changes requested):
Hi, the comments from the previous review still not fixed, you need to make a deploy again and check the demo link after

PR #2461, opened 2025-06-25
langing page

Andriana012:
[DEMO LINK](https://Andriana012>.github.io/layout_landing-page/).


etojeDenys (changes requested):
demo link does not work. Fix it and then re-request the review

Andriana012:
> Демо-посилання не працює. Виправте помилку, а потім надішліть повторний запит на перевірку

Я виправила демо посилання , але коли заходжу на нього все одно не працює так як у мене у браузері . Не показує всі дані

volodymyr-soltys97 (changes requested):
Hi, try to make a deploy again and check the demo link after. Also, if you need some help with your page feel free in the fe_chat

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
image

2. Fix hover styles for this icon and you need to show additional phone number when hovering this icon 
image

3. Add some hover effects to images 
image

4. The height for the inpus looks different as on the design 
image

5. This button should be black by default 
image
image

image

etojeDenys (changes requested):
1. input should not move on hover
image

2. Images should not overlap other objects
image

3. The phone number should only appear on the desktop version and should appear on the left.
example:
image

VitaliyBondarenko1982 (approved):
Looks good

PR #2463, opened 2025-06-26
solution

KazymyrYan:
[demo link](https://kazymyryan.github.io/layout_landing-page/)


KazymyrYan:
I removed animations that are not supported by some browsers.
I just don't understand why my demo link opens only HTML in the browser, without SCSS. On the local host everything is OK (via npm start).

KazymyrYan:
I solved the problem with the demo link

etojeDenys (approved):
🔥

PR #2475, opened 2025-07-01
sol

O1ympus:
[DEMO LINK](https://O1ympus.github.io/layout_landing-page/src/index.html)

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -1,0 +1,43 @@
+# Simple workflow for deploying static content to GitHub Pages
+name: Deploy static content to Pages
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["main"]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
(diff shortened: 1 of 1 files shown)


ArturErmolenko (changes requested):
Add demo link in PR description
Code looks good but i need click Demo

ArturErmolenko (changes requested):
Please double check your work with [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md)

As i see almost all points from check list not implemented
+ pay attention on menu - all links didn't work
also try use tab button in menu - it works strange ;)

src/styles/main.scss
```diff
@@ -1,3 +1,682 @@
...
+    font-size: 16px;
+    line-height: 27px;
+    font-weight: 700;
+    padding: 14.5px 131.5px;
+    margin: 0 auto;
+  }
+}
+
+.contact_us {
+  margin-bottom: 164px;
```
alexander-ignatow, line 396: No margins on the block level

alexander-ignatow (changes requested):
Mucjh better now, now lets work on the details
Please see my comments in the code and fix the following issues:

- Menu exit navigation - text is moving to the left, that looks strange
- Contacts section - no color change for the phone on hover

- Fix the usage for all h1-h6 coponents (see more in comments)
- Fix the usage of a elements - see more in the comments

- remove .css and .css.map files from the repo (use .gitignore file). You are using scss so these files are generated at build time

- break main.scss into separate files. Per BEM - you need a single file er each block.

- for styling - prefer classes with nesting over tags inside classes (unless you have a strong reason to)

PR #2497, opened 2025-07-06
add completed landing

BadOfficer:
[DEMO LINK](https://BadOfficer.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. the phone number should appear when you hover  the icon, and it should be a tel link
image

2. all inputs should be required
image

3. add hover effect to links
image

PR #2505, opened 2025-07-07
Develop

Oleksandr-Oliferuk:
[DEMO LINK](https://Oleksandr-Oliferuk.github.io/layout_landing-page/)


ArturErmolenko (changes requested):
Great job 🔥 
We should fix margins in header and menu on desktop
There should be margins
Header background should have 100% width
Menu also should have 100 % width without any margins
Screenshot 2025-07-17 at 13 20 32

PR #2515, opened 2025-07-08
Developed landing page

vladsulyma:
-  [DEMO LINK](https://vladsulyma.github.io/layout_landing-page/).

### DIFF
--- a/src/styles/blocks/header/top-bar.scss
+++ b/src/styles/blocks/header/top-bar.scss
@@ -10,9 +10,6 @@
     top: 0;
     z-index: 1;
     padding-inline: 0;
-
-    @include on-small-desktop {
-    }
   }
 
   &__logo-link {
--- a/src/styles/utils/mixins.scss
+++ b/src/styles/utils/mixins.scss
@@ -72,7 +72,8 @@
   @include on-tablet {
     gap: $gap-icons-tablet;
   }
-  @include on-large-desktop() {
+
+  @include on-large-desktop {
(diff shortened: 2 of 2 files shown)


src/styles/blocks/page/page.scss
```diff
@@ -0,0 +1,35 @@
...
+  scroll-behavior: smooth;
+
+  &:has(.page__menu:target) {
+    overflow: hidden;
+  }
+
+  &__body {
+    margin: 0;
+    min-width: 320px;
+    background-color: #1d1d1d;
```
alexander-ignatow, line 15: That should be variable

src/styles/blocks/page/compare-bikes.scss
```diff
@@ -0,0 +1,24 @@
+.compare-bikes {
+  &__products {
+    @include page-grid;
+  }
+
+  &__product {
+    grid-column: 1 / -1;
+
+    transition: transform 1s;
```
alexander-ignatow, line 9: IMO - this transition feels too slow. Standard timings for such effects are approx 250-500 mSec (more commonly 300)

alexander-ignatow (approved):
Good and clean job

Mind my comments for the future + 2 more
- It often  makes sence to add some reset or normalize file for such complex projects
- Minor, but it sometimes helpful to create some sort of the `section` bolock - to use it as a base for the sections of the page (for instance set width and horizontal paddings, depending on screen size)

PR #2522, opened 2025-07-09
Develop

MariMajorana:
[DEMO LINK](https://MariMajorana.github.io/layout_landing-page/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -370,11 +370,11 @@
               </div>
             </article>
           </div>
-          <div class="form">
-            <form
-              class="form__inputs"
-              onsubmit="this.reset()"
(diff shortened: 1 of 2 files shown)


pedro-ruas (changes requested):
You did a great job! However, there are a few missing details:
- When hovering the phone icon at page top, the phone number should appear floating next to it
- Images should scale when hovered (you did this), but this scaling should be smooth, with a transition. You can use the property `transition` for that
- All fields in the Contact Form should be **required**, and the page should not reload after the form is submitted
- Some page parts, the form and the first image grid, seem to be off-center. They both should be centered on the page

PR #2532, opened 2025-07-14
layout_landing-page-TetSh

TetiankaSh:


etojeDenys (changes requested):
almost done

1. when hovering inputs it shifts the other staff a little, but it shouldn't. try to hover the inputs in a short period of time

2. these links should lead to Tel, Mailo, and Google Maps addresses
image

3. it should be a Tel link and the user should be able to hover over and click the phone number
image

TetiankaSh:
hello!

Thank you very much for the help and all the comments!

But I'm still not sure I did it right - should I be able to click not only the phone icon, but also the phone number that appears on hover?
For now, I still can't click that phone number, it still disappears when the mouse is out of that phone icon

Do I need to make that phone number stay even when the hover effect is over, so that I can click it as well? Or clicking the icon is enough?

Thanks a lot!

etojeDenys (approved):
Good job!

PR #2546, opened 2025-07-21
Develop

dejucee:
https://dejucee.github.io/layout_landing-page/


etojeDenys (changes requested):
1. remove the link style from the phone number
your:
image

layout: 
image

2. too much space here
image

3. slightly increase size of the images when hovering them

4. add some space here
image

5. too much space here
your: 
image

figma:
image

6. it should be links. The links should be of the correct type
image

7. don't forget to implement these steps:
image

etojeDenys (approved):
Good job

PR #2568, opened 2025-08-05
add task solution

PODDN:
- [DEMO LINK](https://PODDN.github.io/layout_landing-page/).


etojeDenys (changes requested):
1. the phone number should only show up on the desktop version
image

2. The page shouldn't be reloaded on form submit, but the fields should be cleared. Use:
image

PR #2574, opened 2025-08-07
done7

RomaKarpenko:
[DEMO LINK](https://RomaKarpenko.github.io/layout_landing-page/)


FaiHamid (changes requested):
Great job! 🎉
Just few small improvements :

**1.Price should be on one line.**
image
image

**2.Title should be in two lines.**
image
image

**3.Also  gap between section should be smaller.** 
Безntcімені
image
image

**4. The text should remain black after writing, and the font should also match.**
image
image

maxim2310 (changes requested):

1. icons and logo should be in svg format 
image
2. not fixed 
image
3.  check all previous comments, one more time 
4. it should be mailto:
image
5. all external link should open in new tab
image
6. not sure but plz check size and indents of header according design 
image

PR #2590, opened 2025-08-14
solution

Artur2912:
[DEMO LINK](https://Artur2912.github.io/layout_landing-page/)


src/styles/blocks/header.scss
```diff
@@ -0,0 +1,40 @@
+.header {
+  // display: none !important;
```
mkyloa, line 2: let's remove that

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,40 @@
...
+
+    @include on-tablet {
+      padding-bottom: 64px;
+    }
+  }
+
+  &__title {
+    margin: 0;
+    font-weight: 700;
+    color: #ffff;
```
mkyloa, line 27:
```suggestion
    color: #fff;
```

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,36 @@
+.page {
+  font-family: Manrope, Arial, serif;
+  color: $main-text-color;
+  font-size: 16px;
+
+  scroll-behavior: smooth;
+
+  // scrollbar-gutter: stable;
```
mkyloa, line 8: and here

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,36 @@
...
+    min-width: 320px;
+  }
+
+  &__menu {
+    position: fixed;
+    top: 0;
+    left: 0;
+    right: 0;
+
+    transition: all $transform-duration;
```
mkyloa, line 21: do not apply transitions to `all` as it may impact performance and is a generally bad practice

mkyloa: https://stackoverflow.com/questions/8947441/css3-transitions-is-transition-all-slower-than-transition-x

src/styles/blocks/section-title.scss
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  @include on-desktop {
+    font-size: 48px;
+    line-height: 48px;
+  }
+
+  // &section-title--allign-left {
+
+  // }
```
mkyloa, line 21: here as well

src/styles/blocks/header.scss
mkyloa: let's add this to your color variables

mkyloa (changes requested):
all in all awesome work 🚀 
keep it up! 
let's improve on that even further and wdyt about implementing another landing from the selection to go into your portfolio? i'd recommend a more challenging one

PR #2600, opened 2025-08-17
add task solution

Ivadriil:
https://Ivadriil.github.io/layout_landing-page/


etojeDenys (changes requested):
demo link does not work. fix it and then request the review again
 
image

etojeDenys (changes requested):
almost done

1. all inputs should be required
image

2. the title text should be larger
image
image

figma: 
image

image

etojeDenys (approved):
good job

PR #2622, opened 2025-08-26
add task solution

SergeyTsidylo:


etojeDenys (changes requested):
provide a demo link and then request the review again

PR #2623, opened 2025-08-26
solution

nrxyouh:
[DEMO LINK](https://nrxyouh.github.io/layout_landing-page/)


mkyloa (changes requested):
hi 👋 
couple of important fixes from the first sight:
- we should always use svgs for icons. do you know why?
- phone icon size is wrong
- favicon looks wrong

these fixes are a mus

PR #2624, opened 2025-08-26
add task solution

KiT4Er1:
- Deploy and create a Pull Request with a [DEMO LINK](https://KiT4Er1.github.io/layout_landing-page/).


maxim2310 (changes requested):

image

2. align it 
image

3. different fonts here 
image

4. after submit you have to reset form but not refresh page 
5. on mobile first block should be on the whole screen 
image

PR #2625, opened 2025-08-26
Develop

Chips666:
[DEMO LINK](https://Chips666.github.io/layout_landing-page/)


alexander-ignatow (changes requested):
Good start, but please fix a few issues

And double-check if the phone icon size is correct - it feels like it is out of the design
 
image

ArturErmolenko (approved):

PR #2635, opened 2025-09-01
add

yurchornyi:


etojeDenys (changes requested):
great job

1. use images without marks
image

2. fix padding here 
image

3. all inputs should be required

vitalii-bondarenko-ua (approved):
Looks good.

PR #2644, opened 2025-09-05
MyBike landing-page

MarinaKramarchuk:
[DEMO LINK](https://marinakramarchuk.github.io/layout_landing-page/)


src/index.html
```diff
@@ -16,7 +42,387 @@
...
+          </a>
+          <div class="top-bar__icons">
+            <a
+              href="#"
+              class="icon icon--close"
+            ></a>
+          </div>
+        </div>
+        <nav class="menu__bottom">
+          <nav class="nav menu__nav">
```
781calmstate, line 106: The double `nav` looks a little bit off. Usually, we shouldn’t nest `<nav>` inside another `<nav>`. Can we keep the outer one and replace the inner with a `<div>` or just a `<ul>`.

781calmstate (approved):
Great job! The demo is looking fantastic😍 
Fix the issues I mentioned in the comment with the screenshots and it will be just amazing 👍

PR #2646, opened 2025-09-05
Develop

Julia-Hertseva:


mkyloa (changes requested):
please share a demo link
I can't review your work without a demo link :(

mkyloa (changes requested):
still no demolink :( it should be in the pr description

- is this text italic in the design?
image

- the form has extra styles and wrong placeholder styles
image

- the form can be sent with an empty textarea

- hero block on mobile looks different in the design 
image

PR #2649, opened 2025-09-06
Add layout landing page solution

nastiaserkh:
[DEMO LINK](https://nastiaserkh.github.io/layout_landing-page/)

### DIFF
--- a/src/styles/menu.scss
+++ b/src/styles/menu.scss
@@ -54,6 +54,14 @@
     @media screen and (orientation: landscape) {
       overflow: auto;
       max-height: 240px;
+
+      @include on-desktop {
+        max-height: unset;
+      }
+
+      @include on-tablet {
+        max-height: unset;
+      }
     }
   }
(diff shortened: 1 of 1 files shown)


mkyloa (changes requested):
- still extra styles on form elements
image

- page should not be reloaded when i submit the form
- add a favicon
- change the title to better reflect the landing topic
- remove extra styles from this button
image

- the page should not be scrollable when the menu is opened
image

781calmstate (changes requested):
The message field should be required. Currently form allows submission without message field filled.

If the viewport is smaller than the menu, user can't reach some options, so let's add scroll to the menu.
image

781calmstate (changes requested):
Now on the big screens the menu is hidden.
image

781calmstate (changes requested):
image
The menu cropped again if the viewport height is too small

PR #2652, opened 2025-09-07
landing page

MaksimShe:


src/index.html
```diff
@@ -7,16 +7,222 @@
...
+    <link href="https://fonts.googleapis.com/css2?family=Poppins:wght@400;500;600;700&display=swap" rel="stylesheet">
   </head>
   <body>
-    <h1>Landing Page</h1>
+
+    <header class="header">
+      <div class="top-bar">
+        <a class="top-bar__logo" href="#"></a>
+        <div class="top-bar__menu">
+          <div class="header__phone_number">
```
781calmstate, line 24: a bit confusing, is `top-bar` element of `header` block or not? you have `header__phone_number` but its parent `top-bar__menu`

Also usually underscores in bem used to describe the element, so it's better to use `-` instead of single underscore in nameing - `header__phone-number`.

Check your class names by bem for the rest of the html file

781calmstate (changes requested):
The demo link is missing.
Something definitely went wrong with your files, you have too many "copy" files. Consider checking which files you need and which you don't and remove the extra files.

MaksimShe:
https://maksimshe.github.io/layout_landing-page/

mkyloa (changes requested):
- the page should not be scrollable when i open the menu
image

- add a favicon and a page title
- page should not reload when i submit the form
- add a smooth scroll to a page

PR #2653, opened 2025-09-07
[Landing Page - MyBike]

Bohdan-Diatliuk:
[DEMO LINK](https://bohdan-diatliuk.github.io/layout_landing-page/).


781calmstate (commented):
Great job, the demo looks fantastic!

781calmstate (changes requested):
Pay more attention to the checklist for the tasks. You missed some of the checklist points -  point 6, 12, 14, 15.

mkyloa (changes requested):
- use vector images for icons
image

- mobile form looks different in the design
image

- no link to gmaps
image

- typo
image

- weird hover effects and different on two images from the same block
image

- no validation when i try to send an empty form
- wrong font styles and text in placeholders

igorkruz (changes requested):
Looks good, but let’s improve your solution 🙂
- The form on mobile doesn’t look the same as in the design.
- When I press the email button, I should be redirected to the email app to send an email, but it doesn’t work.
- You should add a cursor pointer on the buttons. For the items that scale on hover, it’s also a good idea to add a cursor pointer.

image

mkyloa (approved):
- I SHOULD NOT be redirected to the microsoft website when i submit the form
 - let's add some hover effects here
image

- this button does nothing
image

Overall, great work! 🤝 I approve it, but the issues should definitely be fixed

PR #2656, opened 2025-09-07
landing-page

MaksymBerber:
(https://MaksymBerber.github.io/layout_landing-page/).

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -5,8 +5,11 @@
   class="page"
 >
   <head>
-    <link rel="icon" type="image/png"
-      href="image/unnamed.png">
+    <link
+      rel="icon"
+      type="image/png"
+      href="image/unnamed.png"
+    />
     <meta charset="UTF-8" />
     <meta
       name="viewport"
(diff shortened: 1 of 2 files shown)


mkyloa (changes requested):
demo link is not working, i cannot make a review without a demo link 😞

igorkruz (changes requested):
Good job! 👍
But you should fix the hover effect on your cards and add some padding to the text inside the cards.
There’s also an issue with the hover effect on image cards — you’ll see it in the video.
And also, you should add a transition to make the hover effect smoother.

Video link: https://www.loom.com/share/12f8c5f845f549fd9fced850931aa7ae?sid=f1fe61f4-0b96-4dda-8eb9-881adeac789c

781calmstate (changes requested):
Great job, the demo looks good! 👍 

Let's improve some things:
- The vector images would suit better for icons.

- The address, email and phone number should have hover effects.

- This block should take the whole space according to design.
image

- The `Contact Us` on screens smaller then 1260 should looks like this
image

- The text of cards is overlapping near card's text. Let's add background for text block or some shadows. Your choice, but the text "under" shouldn't be visible.
image

781calmstate (changes requested):
The address, email and phone number still have no hover effects.

PR #2665, opened 2025-09-09
add task solution 1

YaroslavKra:
- Deploy and create a Pull Request with a [DEMO LINK](https://YaroslavKra.github.io/layout_landing-page/).


FaiHamid (changes requested):
Greate job!!
Let`s just fix few details:

1. Change font-size for header title.
image
image

2.As for the images, object-fit didn't help because you uploaded them as squares. Try uploading rectangular images from Figma.
image

3. Add a hover effect to cards and images.
image
image

4. Add more space between text and button.
image

5.On the mobile version, the button should stretch.
image
image

etojeDenys (changes requested):
1. instead of the whole row, you should add a hover effect to the images
image

2. remove base outline from inputs
image

3. all inputs should be required and of the correct type

4. send button should clear the inputs without refreshing the page

5. disable resizing of the text area
image

6. footer is too small
image

7. add space between the icons
image

8. after hovering over this icon on the desktop version, the phone number should appear
image

9. logo should be bigger
image

etojeDenys (approved):
good job. it would be better if the entire text under "address" turned red when hovered
image

PR #2667, opened 2025-09-10
add task solution

webqarbon:
Live Preview: [View Demo](https://webqarbon.github.io/layout_landing-page/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -35,7 +35,7 @@
             </a>
             <div class="header__icons">
               <a
-                href="#"
+                href="tel:+123455555555"
                 class="icon icon--phone"></a>
               <a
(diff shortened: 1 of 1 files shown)


igorkruz (changes requested):
Looks good, but let’s improve your solution:

When I open and close the sidebar, the page jumps because the scrollbar shows and hides — you should fix this.

The phone, email, and address should be clickable and redirect to the correct views. For example:

Clicking the email should open the mail app to send an email.

Clicking the address should open the map with the location.

The button text is also jumping on hover — this should be fixed.

Video link:  https://www.loom.com/share/b1e640d4838a41768e8a7e8e5e34a40b

PR #2678, opened 2025-09-13
Landing page "Nothing"

Anvrey:
- Deploy and create a Pull Request with a [DEMO LINK](https://Anvrey.github.io/layout_landing-page/).

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -55,6 +55,7 @@
             </a>
             <div class="top-bar__icons">
               <div class="icon icon--phone">
+                <a href="tel:+12345555555" class="icon__link"></a>
                 <a href="tel:+12345555555" class="icon__phone-number">
                   +1 234 555-55-55
                 </a>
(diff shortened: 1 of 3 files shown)


etojeDenys (changes requested):
1. increase the size of images when hovering them
2. make sure that all buttons follow this makeup
image

3. fix this section
image

4. these should be tel, mailto and google map address links
image

5. all inputs should be required and of the correct type
6. send button should clear the inputs

7. on the desktop version, the phone number should appear when you hover this icon
image

example:
image

FaiHamid (changes requested):
Greate job!
Let`s just fix a few details:

1.It is worth removing the horizontal scroll.
image

2. Button 'Send' should look like button 'View all'. And form fields should have different borders for different states.
image

vitalii-bondarenko-ua (approved):
Looks good

PR #2685, opened 2025-09-15
solution

whykelly:
- [DEMO LINK](https://artemstadnik.github.io/layout_landing-page/)


etojeDenys (changes requested):
almost done

1. disable text area resizing
image

2. it would be better if this section took up the full height of the viewport
image

etojeDenys (approved):
good job

PR #2686, opened 2025-09-15
add task solution

Bruno-Rauber:
- Deploy and create a Pull Request with a [DEMO LINK](https://Bruno-Rauber.github.io/layout_landing-page/).


raulriato (changes requested):
You're not missing much.

Make sure to add a `cursor: pointer;` to the links and images, so it looks better. Also, the images, they should also have a `hover` increasing effect.

DanielCamposDevX (approved):

PR #2696, opened 2025-09-18
Develop

anattim:


etojeDenys (changes requested):
good job

1. on the desktop version, phone numbers should appear when you hover over this icon
image

example: 
image

2. increase the size of images when hovering them

3. these should be tel links
image

4. all inputs should be required

5.  send button should clear the inputs without refreshing instead of leading to an error page

anattim:
hi
I can't use required for inputs at the same time and have the button clear the fields after submitting
I can't do this without JavaScript here, and I don't know how to do it yet

etojeDenys (approved):
great job. next time, don't forget to add a demo link

PR #2714, opened 2025-09-24
Complete landing page development

ViktoriiaPitsan:


781calmstate (changes requested):
Please provide the demo link, can't review your work without it

781calmstate (changes requested):
Looks good! 😍 Let's make some improvements to make it look even better 👍 

I recommend to use svg for icons instead of png. The svg scales better, easier to style and a good practice for icons.

The email within "Contact Us" section isn't clickable and doesn't have hover effect.

The header text color looks different from design.

Image

The header phone icon should have a hover effect, look for clues in the design to see what it should look like on hover.

Is that me, or it really looks like the left block is a bit lower?

Image

I'd recommend you to keep similar things consistent. For example, I can see that your buttons have transition for hover effects while your phone and address don't have any transition.

PR #2718, opened 2025-09-27
Develop

andriiLink:
[DEMO LINK](https://andriiLink.github.io/layout_landing-page/)


781calmstate (changes requested):
Not bad! 👍  A few fixes and it would be great!

Better to use svg for icons instead of png for better quality.

- The header looks different on the design

Image

- Sidebar header layout is broken

Image

- Something bad happened to your bikes

Image

- It seems like this text could use a little bit more vertical spacing from both sides

Image

- the phone icon in the header should do something on click, it also has to have a hover effect


I suggest you to re-check the checklist and synchronize your landing page with the design before requesting for review.

PR #2724, opened 2025-09-30
Ajustes no estilo do main.scss

Weuler007:
[DEMO LINK](https://weuler007.github.io/layout_landing-page/).


pedro-ruas (changes requested):
There is no code added to the PR, and no DEMO Link either

raulriato (changes requested):
There is still no code in the `pull request`, but yout `DEMO LINK` shows a completely different page than what asked for the task

raulriato (changes requested):
The problem is you opened the `pull request` from your `develop` branch, but you did the task on the `main` branch.

You should merge your `main` branch to your `develop` and then `push` it. Right now there's no way to review your code because it doen't exist on this `pull request`

DanielCamposDevX (changes requested):
You are almost there.
Remember to not have a max-width on the page, on larger screens the page has some white bars on the sides.
On tablet and mobile there are some layout changes (on image grids for example from 3 per row to 1 per row), remember to revalidate the layout on smaller screens.
Remember to place an way to close the site menu, right now if we open the menu we cant close it.

index.html
```diff
@@ -6,17 +6,400 @@
...
+            <div class="exhibition__gallery exhibition__gallery--poster">
+              <a href="#view">
+                <img
+                  class="exhibition__photo"
+                  src="./src/images/Photos 3.png"
+                  alt="Inspiring Walt Disney: The Animation of French Decorative Arts"
+                />
+              </a>
+            </div>
+          </article>
```
raulriato, line 269: These are wrong. The first article is correct, but theses ones you took from the whole section to add here. You should fix it, caus all images are individual and the texts should not be images.

src/styles/main.scss
raulriato, line 1: You should not have a silge file with all the styles. Separate it into at least `blocks` and `utils`

src/styles/main.scss
```diff
@@ -1,3 +1,831 @@
...
+
+  @media (min-width: 744px) {
+    width: min(100% - 144px, 1020px);
+  }
+
+   @media (min-width: 1260px) {
+    max-width: $desktop-container;
+    width: calc(100% - 240px);
+  }
+}
```
raulriato: This task do not require you to use these `min` functions. They are actually causing your layout to break. Consider working only with percentage widths instead.

raulriato (changes requested):
Again, it's a little hard to lacate exactly where the problems in your code are because you left all the styles in one single file. Please make sure to separate it and fix the problems I recommended.

Remember to `deploy` your code after pushing it, so the `DEMO LINK` keeps up with the code.

raulriato (approved):
Well done.

PR #2730, opened 2025-10-01
lp

vladiksanl:


Anton-Kuchmasov (changes requested):
Please deploy your Landing page and then re-request a review
Also, please add a DEMO_LINK to description of your PR
image

vladiksanl:
https://vladiksanl.github.io/layout_landing-page/

etojeDenys (changes requested):
1. add some space here. also, the phone number should be tel link as well
image

2. add smooth scrolling to the page

3. make sure the menu looks the same as the layout
your:
image

figma:
image


4. increase the size of images when hovering them

5. download the image without the mark
image

6. all buttons should follow this mockup
image


7. these should be links with correct type
image

8. all inputs should be required

9. add some space here
image

10. make sure that this section looks the same as the layout
your:
image

figma:
image

PR #2733, opened 2025-10-03
add solution

andreysaltan24:
[DEMO LINK](https://andreysaltan24.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. add a hover effect to the header items

image

example:
image

2. these links do not work
image

3. install the image without the mark
image

4. decrease the scale of the hovered image a little bit
image

5. after clicking on the input, its size changes, but it should not

PR #2734, opened 2025-10-04
landing page

VitaliyHoroshko:
Hope it's good


etojeDenys (changes requested):
1. logo and links should be on the same line
image

2. menu logo should be a different color

3. add a hover effect to the links. also, those links should work
image

example:
image


4. these should be links with the correct type
image

5. make sure this section looks the same as the layout
your:
image

figma:
image

6. all of your buttons should  follow this mockup
image

7. should be tel link
image

etojeDenys (changes requested):
almost done
1. these should be links to the following: tel, mailto, google maps address
image

2. icons look too small on the desktop version
image

3. there is too much space here
image

4. the phone number should only show up on the desktop version
image

etojeDenys (approved):
good job

PR #2737, opened 2025-10-07
Landing Page: exercise of mate academy

Alex-Rw2:


pedro-ruas (changes requested):
There are some issues with your project and also some missing required features:
- Header section:
  - Header styling is wrong, the Menu item should be located to the right of the screen, and the page logo to the left, not in the middle of it. (Header should be full width)
  - When hovering the phone icon at page top, the phone number should appear floating besides it
  - Initial page banner and menu should both occupy full page height (100vh)
  - Banner text is not correctly positioned
  
- Images section:
  - Images should have a scale-up animation when hovered
  - Images section is missing correct styling, some images are smaller than they should
  - Spacing and fonts do not match the original design in Images section

- Contact form section:
  - All fields in the contact form should be required, including the message
  - When sending the form, the page should not reload, but the form should clear itself
  - Address, phone number, and email, should all be links to their respective targets
  - Styling for the form fields placeholders does not match designs 

You should also run `npm run deploy` to deploy your code, and add the [DEMO LINK](https://alex-rw2.github.io/layout_landing-page) to the PR

src/styles/main.scss
```diff
@@ -1,3 +1,18 @@
 body {
+  background-color: $color-page;
   margin: 0;
 }
+
+* {
+  border: 1px solid red;
```
pedro-ruas: Why?

pedro-ruas (changes requested):
You did not implement every request from my last review.
You also forgot to run `npm run deploy`, as asked.
There is a border around everything that should not be there

src/styles/main.scss
Alex-Rw2: my apologies I'm doing this project on 2 computers at the same time.


-* {border: 1px solid red;}
-This is for better visibility of the page blocks. My mistake, I forgot to delete it.

"npm run deploy" apparently doesn't change my project at all... in my eyes it's not important that's why I always forget to run it but I'm going to release a commit right now running this command I've forgiven myself again

(And i use google translator for some words so forgive to some writing incompatibility)

Alex-Rw2: If possible, give me some guidance on what is missing because I don't know if I fully understand what to do or how to do it...

pedro-ruas (changes requested):
The command `npm run deploy` is for making your project available for our testing online, without it we mentors need to clone the project and install locally for testing. Because we have to review several projects daily, this takes an unnecessary time as the deploy is a part of the task.

Now about what is missing:
- The contact form is clearing when sent and the page isn't refreshing. Great, but the fields should all be required too. (Just add "required" to the tags)
- Email, address and phone beside the form, should be links, not just plain text. `<a href="tel+..."` for example. Phone link should start a call, address should direct to maps and email should start an email
- When hovering the phone icon at page top, the phone number should float beside it
- The menu still isn't taking the full page height when open
- The font-family for the first paragraph (under the main banner) is wrong, looks different from everything else in the page
- The hovering effect on images should be on all the images in images section, not just on the first 3. And the effect should be on the image only, not in the text under it

If you need help fixing any of those, just ask in the questions chat in the platform and we will help you

pedro-ruas (approved):

PR #2745, opened 2025-10-12
Develop

nols212:
[DEMO LINK] (https://nols212.github.io/layout_landing-page/).


etojeDenys (changes requested):
1. increase the size of images when hovering them

2. these should be links with the correct type (telephone, email, google maps address)
image

3. disable text area resizing
image

4. do not forget to implement these steps
image

5. phone number should appear when you hover over this icon
image

etojeDenys (changes requested):
almost done

1. on the desktop version, the phone number should appear when you hover this icon
image

example:
image

2. the inputs and text area should be slightly larger
image

example:
image

etojeDenys (approved):
good job
however, the phone number should only show up on the desktop version
image

PR #2755, opened 2025-10-17
Solution 1

ImPEPEGA:
[DEMO LINK](https://ImPEPEGA.github.io/layout_landing-page/)


Anton-Kuchmasov (changes requested):
Good job!

However, there are a few issues need to be fixed:

1. Add an ability to click on numbers (use `tel:` with `<a>`):
image

2. Check styles here:
image


3. On mobile devices (phone, tablet) with sensor-screens, there should be no hover-effect on phone icon:
image

Anton-Kuchmasov (approved):

PR #2756, opened 2025-10-17
completed tasks

qlaudy:
(https://qlaudy.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. slightly decrease the scaling so that images do not overlap other items
image

2. these should be links with correct type. also, there is too much space between them
image

3. add some space there and disable text area resizing
image

4. on the desktop version, the phone number should appear when you hover this icon
image

qlaudy (commented):
(https://qlaudy.github.io/layout_landing-page/)

qlaudy (commented):
[DEMO LINK](https://qlaudy.github.io/layout_landing-page/).

FaiHamid (changes requested):
Good job!
Let`s just fix a few things:

1. Add hover effect to elements in the recommended section
image

2.I can still change the size of the textarea; you need to add the `resize: none` property to this element.
image

3.The icons are not visible on a large screen because they blend in with the background image.
image

4. Still too much space between contacts
image

5.Also, on mobile devices, both elements in the contacts section should be stretched across two columns.
image

qlaudy (commented):
(https://qlaudy.github.io/layout_landing-page/)

etojeDenys (approved):
great job! just remove the background color from those icons
image

PR #2763, opened 2025-10-20
add solution

77carlos71:
- Deploy and create a Pull Request with a [DEMO LINK](https://77carlos71.github.io/layout_landing-page/).


pedro-ruas (changes requested):
Carlos, there are a few issues that you need to address:
- The Menu should cover the whole page both in width and height, so there is no scroll when it's open,
- The Banner at the page beginning and the footer should both have full screen width
- The images at "Now on view" should have the hovering effect, similar to what you did for the Met Logo at page top
- The "More to explore", "Collection", and Contact sections are broken, images are overlapping text and no following the style. These images should also have the hovering effect
- When sending the contact form, it should clear itself, but not reload the page. use a function for that
- Phone numbers and addresses aside the contact form should all be links, phone link and google maps link respectively

PR #2766, opened 2025-10-21
add task solution

romandrabyniuk:
- [DEMO LINK](https://romandrabyniuk.github.io/layout_landing-page/)


FaiHamid (changes requested):
Greate work!
Let`s just fix a few details:

1.You need to remove the horizontal scroll bar.
image

2.Nothing happens when I click on the email in my contacts.
image

Anton-Kuchmasov (changes requested):
Please get rid of horizontal scroll:
image


Also, would you mind to make phone number here clickable?
image

etojeDenys (changes requested):
almost done! here are the last fixes

1. number is clickable now, but you should also add a cursor pointer
image

2. phone number should only show up on the desktop version
image

etojeDenys (approved):
good job

PR #2777, opened 2025-10-27
Develop

Evg3n22:
- Deploy and create a Pull Request with a [DEMO LINK](https://Evg3n22.github.io/layout_landing-page/).


etojeDenys (changes requested):
1. add a hover effect to the menu links
example:
image

2. add a hover effect to the header items
image
example:
image

3. increase images on hover

4. these should be links (tel and addresses)
image

5. disable textarea resizing
image

6. do not forget to implement these steps
image

7. fonts in this section look different
image

etojeDenys (changes requested):
1. images should not overlap other items. decrease the scale slightly
image

2. these should be links to the addresses on google maps
image

3. all inputs should be required, and the "send" button should clear the inputs

FaiHamid (changes requested):
Good job!
Just needed a few fixes: 

1. Font looks different than in figma.
image
image

2. Text with contacts should appear after hovering phone icon, not all top header part.
image

3.After clicking 'send' page reload, but shouldn`t do it.

etojeDenys (changes requested):
almost done

1. after filling out the fields and pressing the send button, the page scrolls up, but the inputs are not cleared

2. these links should open in a new tab
image

3. user should be able to hover and click on these numbers
image

FaiHamid (changes requested):
Good job!
You almost done it, just one last fix:

1.Add hover for icons (For example, you can slightly increase their size when hovering over them):
image
image

etojeDenys (approved):
good job!

PR #2782, opened 2025-10-29
solution

yarovskiy-a:
https://yarovskiy-a.github.io/layout_landing-page/

✅Use the logo as a favicon
✅Use the landing name as a page title
✅All logos should be links to the top of the page
✅Change text color on hover for phone, email, and address
✅Make sure phone icons and phone numbers are real links to start a call
✅All addresses should be links to some location on Google Maps. Open them in a new tab with target="_blank".
✅Increase all images on hover (make them links to #, if there is no better option)
✅Make sure everything looks neat on mobile and without horizontal scrolling
✅The speed of animations should be the same throughout the page (for example, increasing when hovering or moving blocks when scrolling)
✅Ensure that all nav__links work smoothly
✅Disable page scrolling under the menu using the next code:
}
✅Make it possible to scroll the menu if it is higher than the viewport;
.menu {
}
.menu__top {
  top: 0;
  z-index: 1;
  background-color: #e5e5e5;
}
✅All form fields should be required and have correct types (email, tel, etc.)
✅Placeholders should give examples of what should be added, to make the expected format clear
✅The page shouldn't be reloaded on form submit, but the fields should be cleared. Use:
✅Make the footer sticky at the bottom with z-index: -1 to get a fixed background effect.


volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. Add a favicon to the page and change title to `My Bike`
image

2. You need to show additional phone number when hovering this icon, check it on the design 
image

image

4. This section looks different as on the design 
image

Design
image

5. The images shouldn't overlap each other when hovering 
image

6. Looks different, fix all styles as on the design 
image

yarovskiy-a:
✅Added an icon.

✅Fixed the problem with the phone number.

✅Aadded effect hover for all clickable elements.

✅Design has been fixed.

✅Fixed image overlap each other when hovering.

✅Fixed the design of the contacts section.

etojeDenys (changes requested):
1. use higher-quality images for the desktop version
image

2. download the image without the mark
image

3. remove base outline
image

4. all inputs should be required and have the correct type

5.  send button should clear the fields

yarovskiy-a:
✅ Used a higher-quality images for the desktop version.

✅ Сorrected the image.

✅ Base outline removed.

✅ All input data is required and has the correct type.

✅ Send button clears fields.

etojeDenys (approved):
great job

PR #2785, opened 2025-10-31
add task solution

nnovik:
[DEMO LINK](https://nnovik.github.io/layout_landing-page/)


maxim2310 (changes requested):
Good job! Let's improve your landing a bit)


1. add hover effect on interactive elements 

Image

2. submit button should just reset form, not refresh page 

Image

3. these should be links 

Image

4. add fav icon

h-amster (approved):
Good job! You can add scale on hover to images from screenshot as well.

image

PR #2795, opened 2025-11-04
add task solution

deny-hl:
- Прев'ю: [https://deny-hl.github.io/layout_landing-page/](https://deny-hl.github.io/layout_landing-page/)


h-amster (changes requested):
1. Phone number should be visible on icon hover
image

image

2. Export image without label

image

3. Form implementation differs from the design.

- Remove default outline on focus.
- Set correct font for placeholder
- Set correct paddings for input
- The page shouldn't be reloaded on form submit, but the fields should be cleared

image

4. Add some behavior for all interactive elements on hover, like scale for images and text color change for links

5. Make the footer sticky at the bottom with z-index: -1 to get a fixed background effect.

brespect (changes requested):
Good work! 
1. Don`t forger to resolve conflicts before submitting
2. Looks like not responsive form inputs:
{9A0896EB-0051-4CE5-A4C2-FADCE702D143}
3. Same way main picture look different from what expected: 
{ACC3D4AD-0D2E-4DD8-AE8B-175666BBEADB}

4. Let`s look how the links hover effect should looks like:
{B96CF45B-E4AA-4D87-8C42-7CC1FEF9B131}

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. This section looks different as on the design, need to fix it 
image

2. The font color should be are same 
image

image

image

image

etojeDenys (approved):
good job

PR #2799, opened 2025-11-05
solution added

Olena-Dolna:
[DEMO LINK](https://Olena-Dolna.github.io/layout_landing-page/)
[x] Change text color on hover for phone, email, and address
[x] All addresses should be links to some location on Google Maps. Open them in a new tab with target="_blank".
[x] Ensure that all nav__links work smoothly
[x] Disable page scrolling under the menu using the next code:
}
.menu {
}
.menu__top {
  top: 0;
  z-index: 1;
  background-color: #e5e5e5;
}
[x] All form fields should be required and have correct types (email, tel, etc.)
[x] The page shouldn't be reloaded on form submit, but the fields should be cleared. Use:
[x] Make the footer sticky at the bottom with z-index: -1 to get a fixed background effect.


brespect (changes requested):
Good job, almost done!
1. Let`s check alignment of this text on mockup:
{A6457C1F-522E-4C54-B54C-C1FBBFE9A7A8}

2. Let`s add some hover effect to all pictures 
{FE182FB1-6625-45CB-BB5D-B51F5EC2F76E}

3. Look at the mobile main photo, it should looks different on 320px screen
{F8853C31-BF64-4B5C-8037-3180AE2F6335}

brespect (approved):
Well done!

PR #2802, opened 2025-11-07
add landing page

Vladislav-Korsun:


etojeDenys (changes requested):
please provide a demo link and request a review again

etojeDenys (changes requested):
1. on the desktop version, the phone icon should have this hover effect
image

2. make sure these links work and lead to the correct sections
image

3. slightly increase the size of the images when hovering them

4. these should be tel, mailto and address links
image

5. send button should not reload the page. it should only clear the inputs

6. disable text area resizing

h-amster (changes requested):
Right now, your demo link still shows the old version of the project. Please update your deployment so it shows the latest changes, and then request a new review. The link currently opens the version of the site without your recent fixes.

Vladislav-Korsun:
I tried to update the demo link, but I'm not sure the deployment was refreshed correctly. Could you please check it? Also, could you tell me how I can ensure the demo link always shows the latest version of my project?

etojeDenys (changes requested):
1. number should be a link. also, the font is different from the mockup
image

2. these links still do not work
image

3. the send button should not reload the page
image

Anton-Kuchmasov (changes requested):
Good job, but there are a few issues need to be fixed:

1. Remove horizontal page' scroll. Double-check, if there are any big elements (with big width) that stretches your page:

2. Check out font of "Bring joy back" section - it looks weird:
image

3. Click on every link in `aside` menu should scroll page to selected section:
image

4. When `aside` is opened, no scroll should be visible. Check out #11 of [this instruction](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md):
image

5. Would you mind sligthly reduce hover-effect here?
image

6. Add `target=_blank` to this link:
image

vitalii-bondarenko-ua (approved):
Lookd good

PR #2828, opened 2025-11-21
Solution

UArtem87:


FaiHamid (changes requested):
Hi, 
Please, add Demo_link to pull request description.

UArtem87:
Hi, I tried so hard that I forgot
https://uartem87.github.io/layout_landing-page/

brespect (changes requested):
Good progress, let`s make some changes:
1. I can`t submit the form.
2. Buttons should have `cursor:pointer` on hover & should be different by height:
Знімок екрана 2025-11-22 о 11 00 09

3. Add hover transition effect to all images: 
Знімок екрана 2025-11-22 о 11 01 39

4. Why "Cart(0)"?
Знімок екрана 2025-11-22 о 11 02 22

UArtem87:
I think this is correct
https://uartem87.github.io/layout_landing-page/

Anton-Kuchmasov (changes requested):
Good job!

To improve:
1. Check out [autofill ](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:autofill)styles here. 
Make them transparent - here's link for solution that you can use as a reference:
https://romkas228.github.io/layout_landing-page/

image

2. When User sends a form, there should be no scroll-up's, no any other actions but wiping out an entered info
Check out #15 of [this instruction](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md)

3. Each of those links should be clickable - f.e., phone number should trigger phone call (use `<a href="..." tel="+38099..." />`, for address it should open a Google Maps in new tab etc. 
image

4. This section looks different rather than Figma mockup (I mean, background image is seems to be broken)
image

5. Would you mind to slightly reduce hover effect of this cards? They are overlapping each other:
image

Make these changes, re-deploy your solution & request mentor's review again

Anton-Kuchmasov (changes requested):
Good job, but links in Contact Us section are still unclickable:
image

Please add special attributes and re-deploy your solution

UArtem87:
I forgot about them

Anton-Kuchmasov (approved):
Well done!



Add line break (`<br />`) here
image

PR #2841, opened 2025-11-26
add task solution

Vasylysa1333:
[DEMO LINK](https://Vasylysa1333.github.io/layout_landing-page/)


etojeDenys (changes requested):
1. phone number should only appear when the user hovers the phone icon. also, that icon should be a tel link
image

2. decrease the scale of the images to avoid overlap
image

3. that should be a google maps address link
image

4. all inputs should be required

5. send button should clear the inputs

6. add favicon

7. remove scrolling when menu is open (look at checklist, point 11)
image

FaiHamid (changes requested):
Well done 👏
Let`s just fix a one detail:

1.Rectangle images still have large hover effect, and don`t have border-radius.
image

etojeDenys (approved):
good job

however, the phone number should only show up on the desktop version
image

PR #2848, opened 2025-12-04
Dev

Dimononon:
[DEMO LINK](https://dimononon.github.io/layout_landing-page/).

Used a different approach for page paddings and margins between blocks. Instead of using out of BEM .container, all needed changes was implemented with elements of .page block and mixed along with corresponding blocks.


ArturErmolenko (approved):
Hreat job 🔥

PR #2855, opened 2025-12-05
add task solution

DemonWayne:
Implementing a [B&O mockup](https://www.figma.com/file/DtkQmQ797hk0nI4KfMi2Uq/BOSE-New-Version?type=design&node-id=6817-212&t=ZTV6Gl8NzaWkJ4FK-0) layout design using BEM methodology.
On the technical side, HTML with SCSS was used.
Almost all hard-coded values were extracted to a variables file for faster modifications.

[Demo Link](https://demonwayne.github.io/layout_landing-page/)


mkyloa (changes requested):
nice one 👍 

Image

- icon size on mobile is wrong

Image

PR #2857, opened 2025-12-05
Initial commit landing page lemishko

NastyBaster:
https://nastybaster.github.io/layout_landing-page/#


mkyloa (changes requested):
- hero block looks different in the design
Image

Image

PR #2858, opened 2025-12-05
add page for THE MET and readme file

nastia7408:


etojeDenys (changes requested):
1. menu looks broken
image

2. decrease the image scale a little when it hovered because now it looks like it merges with the other image
image

3. use higher quality images for the desktop version

4. fix problem with images here
image

5. disable text area resizing

6. make sure this part looks good in smaller versions
image
image

7. phone numbers should only show up on the desktop version
image

8. these links should open in a new tab
image

etojeDenys (changes requested):
almost done

1. decrease the scale of the image when it hovered. make sure the images do not touch or overlap other items
image

2. use a higher quality image for the logo
image

3. disable text area resizing
image

etojeDenys (approved):
good job. just fix size of the logo
image

PR #2861, opened 2025-12-07
added solution

TetianaSobolieva:
Посилання на прев’ю

https://TetianaSobolieva.github.io/layout_landing-page/

⸻

Посилання на макет

https://www.figma.com/file/DtkQmQ797hk0nI4KfMi2Uq/BOSE-New-Version?type=design&node-id=6802-139&t=L7eKz5YKLN0m5WxR-0


mkyloa (changes requested):
- hover effects should be fixed

Image

Image

- form should be validated

Image

781calmstate (changes requested):
- consider handling this edge case, for now the message input can be resized horizontally which allows user to break your layout and even create horizontal scroll. i recommend to remove horizontal scroll and allow only the vertical one.
Image

- it should be full-width according to the design
Image

- the header text breaks differently from design on some resolutions(not only on mobile), make sure to handle it properly
Image

make sure to complete the points of the checklist.
the next points are not implemented or implemented incorrectly:
`(4) - in the menu the phone color doesn't change on hover, (13) - the message field isn't required, (14)`

dothearts (approved):
Many thanks for the changes! 
Approved!

781calmstate (approved):
Make sure to clean-up your codebase - remove unused files, images etc. and make sure you don't have any unnecessary comments in your code

PR #2862, opened 2025-12-07
Develop

Kanezoor:
https://kanezoor.github.io/layout_landing-page/


2pasha (changes requested):
great job! 👏

please take a look on these improvements: 

1. these elements should be in svg format
image

image

3. please add favicon

4. after form submitting - user have received 450 error. Instead of error user should see imitation of form submitting (reset all fields and do not refresh the page) 
image

5. here should be link to google maps
image

etojeDenys (changes requested):
almost done

1. phone number style looks different with the layout
image

figma:
image

2. phone number should only show up on the desktop version

etojeDenys (approved):
good job

PR #2865, opened 2025-12-08
developing

marcellealvb:
https://marcellealvb.github.io.com/layout_landing-page/


IanLuan (changes requested):
Your landing page is very good and your code looks great. But there are some missing items from the check list that you should take a look:

- Use the logo as a favicon

Remember to deploy your page after the changes.

pedro-ruas (approved):

raulriato (approved):

PR #2867, opened 2025-12-08
Develop

Nazarii-Lesniak:
The background is not displayed on gh pages, I don't understand why. I also couldn't stretch it to the full width of the screen.


Anton-Kuchmasov (changes requested):
Good work you did!

1. To fix an issue with no-background, try to rename your background images (I mean, filenames of them) to be something without whitespaces:
image


after that make, as usual add-commit-push and deploy again

2. Check out styles of [autofill ](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:autofill)here - they should be the same as usual input filling:
image


Also, I can't enter >= 20 chars into `email` field. That's uncorrect - email length can be much longer then 20 chars :)
image


3. Would you mind to add some inline-paddings here?
image

4. User should be able to click on the phone number:
image

2pasha (changes requested):
good job! 👏

please take a look on this improvements:

1. because of this shadows this block wants either inline paddings or remove this shadows like in figma. You need to choose
image

2. all your images have this shadows. As for me it looks little bit out of design system. I would remove this shadows. 
image

3. on tablet this elements should look different
image

4. this text should be bold
image

etojeDenys (approved):
lgtm

PR #2870, opened 2025-12-08
Develop Landing Page

torianykhanna:
[DEMO LINK](https://torianykhanna.github.io/layout_landing-page/).


781calmstate (changes requested):
The code looks pretty good! 👍 

But your demo links isn't working, make sure to fix it, so mentors can see your cool web-site 😍 

I'd also recommend to try and import icons in `svg` format from figma and try to use them instead. SVG format is better for icons and a common practice for all projects

PR #2871, opened 2025-12-08
page

akhvarastyan-web:
https://akhvarastyan-web.github.io/layout_landing-page/


781calmstate (changes requested):
Image
Your header is blank.

Also it looks like some of other images are missing from your web-site.

Before re-requesting review also make sure to check that your landing page looks exactly as on design on figma, because from what I can see there are a lot of differences on your web-site.

I'd also recommend to use svg format for icons. You can choose image format while importing images from figma

dothearts (changes requested):
image

Also blank on your demo link (first screen). 
Please fix these issues, redeploy the project, and ask review. 
Many thanks!

PR #2872, opened 2025-12-09
created project landing page

Etwaiz:
[DEMO LINK](https://Etwaiz.github.io/layout_landing-page/)


781calmstate (changes requested):
The best format for icons is svg, so I'd suggest to try and import those images from figma in svg format.

Make sure to sync your functionality with points from checklist. The next points are not implemented: `(1)`, `(4)`, `(6)`, `(7)`, `(14)`, `(11)`, `(12)`.

Your header text breaks not as it should according to design. I noticed the differences on big screens(1300+) and on small screen (~400px and less). But make sure to check other resolutions.

Image

Before re-requesting the review, I suggest you to check and sync your web-site with the design, so it looks the same.

ArturErmolenko (approved):

PR #2873, opened 2025-12-09
landing page

lilp3ka:
[DEMO LINK](https://lilp3ka.github.io/layout_landing-page/).


781calmstate (changes requested):
Svg format is better for icons, so I'd suggest to try and import images in svg format from figma.

You missed some of the checklist's points, so make sure to sync your work with the list before submitting it. The next points are not implemented or working incorrectly:  
`(1), (4), (5) - icon in the header doesn't start a call + contact us section, (12), (13), (14), (15) - submit button behavior is wrong`

Also work on the issues illustrated below. Also make sure to provide some space for text and placeholders in the inputs.

Image

Image

Before re-requesting review make sure to sync with the design and make your website to look as similar as possible

dothearts (changes requested):
I don't see changes after the previous review. 
Please make it, and fix all issues which described above. 
Many thanks!

igorkruz (changes requested):
You should fix all the suggestions mentioned above.

In the Recommended section, the cards zoom on hover, which creates an interactive effect. It would be better to add cursor: pointer to indicate that the cards are interactive.

Also, the images in the section below do not match the design. Please update them to match the layout and proportions shown in the mockup.

Current implementation:
Image

Expected design:
Image

dothearts (changes requested):
Great job. Fix some issues which i found: 
1. For some rason image in footer has opacity or cover effect.

Image

I don't find this in design, but think we can make this effect lower. (not critical, my opinion) 

2. Fix zoom of items. This cards zoom too big. 

Image

Other - great. Fix this minor issues, make request. 

Many thanks!

dothearts (approved):

PR #2876, opened 2025-12-10
Finished Landing page

Renata11111-afk:
[DEMO LINK](https://Renata11111-afk.github.io/layout_landing-page/)


781calmstate (changes requested):
Svg is a better format for icons, so I'd suggest to import icons in svg format from figma.

- The header text seems to be much lower that it is on the design.
Image

- On smaller screens there is no space between title and content in this block
Image

- These bikes should have hover effect according to the checklist. Also they seems to be a little bit cut in the bottom.
Image

- The phone icon in the header should show the number on hover according to the design
- the submit button in the "contact us" scroll behavior is weird

Make sure to complete all the points from checklist. The next points are missing:
`(1) - wrong image, (4), (5) - contact us section, (6) - contact us section, (7) - compare bikes section, (11), (12), (14), (15), (16)`

Moroz-Dmytro (changes requested):
1. Hover effect on phone icon doesn't look as expected
2. Open menu and scroll 🫨 

Image

Renata11111-afk:
@Moroz-Dmytro thank you for your comment, somehow i didn't notice that before(
I tried my best to fix a srcoll problem without using js.

Moroz-Dmytro (approved):
I'll approve it, but try to fix next comment

When on mobile I open menu and rotate phone -> I don't see all items in list(not possible to scroll them)

Image

PR #2877, opened 2025-12-10
Develop

Demtar23:
https://demtar23.github.io/layout_landing-page/


781calmstate (changes requested):
I recommend using svg format for icons, you can choose the format of image while importing it from figma.

I think you should visit the checklist and make sure to complete all the points. Because most of them are not implemented.

Before re-requesting review make sure you completed all the points from the checklist and your web-site looks as similar as possible to design

mkyloa (approved):

PR #2890, opened 2025-12-17
add task solution

nxqar:
– [DEMO LINK](https://nxqar.github.io/layout_landing-page/).


Anton-Kuchmasov (changes requested):
Great work you did!

To improve:
1. Would you mind to reduce `hover` effect to avoid photos overlapping?
image

2. When user clicks `Send` button, this form should only be cleared - without page reload or smth else. Check out #15 of [this instruction](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md)
image

3. Hover on `phone icon` in Header should show a clickable phone number. 
Check out [this solution](https://nkoropka.github.io/layout_landing-page/) to get clarification (also you can use it as a reference) 
image

brespect (changes requested):
Almost done:
1. Check the icons onMonile version:
Знімок екрана 2025-12-18 о 14 05 49

2. Add z-index property to your Footer Image.

Anton-Kuchmasov (approved):
Great job!

PR #2892, opened 2025-12-17
THE MET Landing Page 

NazarenkoAnna:
[DEMO LINK](https://NazarenkoAnna.github.io/layout_landing-page/). 
[DESIGN LINK](https://www.figma.com/file/lSR1m42L9YwzQwzzxKwHpw/THE-MET)


Anton-Kuchmasov (changes requested):
Good job!

To improve:
1. Would you mind to add `hover` effect on each of those cards? `transform: scale(1.03)` will be ok)
image

2. Font size of `h1` should be 80px according to Figma:
image

But your solution's `h1` now has `64px` font size:
image

Check out all of this moments around a Landing page

3. When user clicks on each link in `aside` menu, the page scroll down instantly. Please make this transition smoother

brespect (changes requested):
Almost done: 
1. Something strange happens when I click on this phone numbers:
Знімок екрана 2025-12-18 о 14 44 27

2. The input text should be same font style:
Знімок екрана 2025-12-18 о 14 46 28

3. Add some scale on hover for buttons

NazarenkoAnna:
> 3. Add some scale on hover for buttons

@brespect Thanks for your review!
I've fixed first two issues. But according to the design, the button only changes background color on hover. We already have scale  on images and I'm think that adding this effect to the buttons would be too much .

Anton-Kuchmasov (approved):
Amazing!

PR #2894, opened 2025-12-18
Develop

NemH:
- [DEMO LINK](https://nemh.github.io/layout_landing-page/).


maxim2310 (commented):

1. scroll should be hidden when menu is open 

Image

2. here should be phone number when i hover on icon 

Image

3. icons and logo should be in svg format 

Image

4. align prices and add some hover effect on cards 

Image

5. animation works only on mouse enter but not on leave (looks like roughs animation )

Image

Image

6.it has to be mailto

Image

7. font is different here 

Image

8. on mobile first screen looks diferent from design 

Image

9.

maxim2310 (changes requested):
.

NemH:
1) 1)  [X] OK
2) 2)  [X] OK
3) 3)  [X] OK
4) 4)  [X]  I didn't see this on the landing page, but I changed it.
5) 5)  [X] OK
6) 6)  [X] OK
7) 7)  [X] OK
8) If the size is specified as 320 +- 20 on the layout, everything looks good, when you zoom in, it starts to look smaller and smaller until it reaches the tablet 576px, I don't know how to fix this, reduce the minimum size of the tablet window or create another version of the page for on-phone-wide?

vitalii-bondarenko-ua (approved):
Looks good.

PR #2907, opened 2025-12-26
Develop

nvkoff:
[DEMO LINK](https://nvkoff.github.io/layout_landing-page/)


Anton-Kuchmasov (changes requested):
Good job!

To improve:

1. When User adds some value as an autofill (chosing one of purposed browser suggestions), styles of [autofill ](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:autofill) should be the same as usual:
image

2. Clicking on `Send` button should ONLY wipe all entered data - no page reloading or other side effects allowed. Check out #15 of [this instruction](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md) to proceed.

3. Would you mind to add some `hover` effect on those cards?
image


`transform: scale(1.04)` will be pretty good option

4. Background positioning on mobile & tablet view seems slightly broken:
image


You can use this solution as a reference (if you're not confident about some styles, you can observe them in `devtools`:
https://gallik-dev.github.io/layout_landing-page/

vitalii-bondarenko-ua (changes requested):
Good job.
Needs improvements.
- prevent resize textarea
Screenshot 2025-12-27 at 08 42 28

- Pay attention on [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md).  You need to implement all points.
For example these points not implemented, but check all.
1. Use the logo as a favicon
3. Change text color on hover for phone, email, and address
...
...

11. Disable page scrolling under the menu.
Screenshot 2025-12-27 at 08 41 31
...
15. The page shouldn't be reloaded on form submit, but the fields should be cleared
16. Make the footer sticky at the bottom with z-index: -1 to get a fixed background effect.

Anton-Kuchmasov (changes requested):
Hey, would you mind to redeploy your solution?

Because all issues that were mentioned above hasn't implemented yet
image

vitalii-bondarenko-ua (approved):
Looks good

PR #2908, opened 2025-12-27
Develop

Lazarin123:
- When done check yourself using the [CHECKLIST](./checklist.md) when finished.
- Deploy and create a Pull Request with a [DEMO LINK](https://Lazarin123.github.io/layout_landing-page/).


IanLuan (changes requested):
Hi Samuel! Your code looks good, but your DEMO LINK isn’t working. Please update it so we can finish the review.

index.html
```diff
@@ -9,14 +12,403 @@
...
+              ></a>
+            </div>
+          </div>
+          <div class="header__bottom">
+            <div class="header__waves"></div>
+            <h1 class="header__title">Take the Streets</h1>
+          </div>
+        </div>
+      </div>
+    </header>
```
raulriato, line 73: Your `header` is missing both the menu and the background image.

image

index.html
```diff
@@ -9,14 +12,403 @@
...
+                />
+                <h3 class="product-title">Agile ride 3</h3>
+                <p class="product__category">
+                  The lightweight frame that has earned its street tread as a
+                  sleek, agile ride.
+                </p>
+                <p class="product__price">$ 2 090</p>
+              </article>
+            </div>
+          </section>
```
raulriato, line 222: although it doesn't need to be pixel perfect, it should at least be well aligned.

image

Make the sections have the same `width`

index.html
```diff
@@ -9,14 +12,403 @@
...
+                  Send
+                </button>
+              </form>
+              <div class="contact-us__informations">
+                <section>
+                  <article>
+                    <p class="contact-us__informations__title">Phone</p>
+                    <a
+                      class="contact-us__informations__contact"
+                      href="tel=+12345555-55-55"
```
raulriato: This is not the way to `href` a link to make a call. Instead of `=`  you should use `:#`

src/styles/blocks/products.scss
```diff
@@ -0,0 +1,35 @@
...
+    @include on-desktop {
+      grid-column: 6 / 12;
+    }
+  }
+}
+
+.product {
+  &__photo {
+    width: 100%;
+  }
```
raulriato, line 31: It would be smoother if you add a `cursor: pointer;` to all the images that represent something 'clickable' on the project

src/styles/blocks/contact-us.scss
```diff
@@ -0,0 +1,107 @@
...
+        line-height: 140%;
+      }
+
+      &__contact {
+        text-decoration: none;
+        font-size: 16px;
+        font-weight: 500;
+        font-style: normal;
+        color: #fff;
+      }
```
raulriato, line 116: image
you only reseted the decoration on desktop. It should be this way in any device.

raulriato (changes requested):
Your project seems okay, but have a few critical problems that you should fix.

DanielCamposDevX (approved):
Well done! Everything seems good just some minor changes that could improve your project (like we discussed on the Q&A):

- Becarefull with padding in mobile screens
- The effect on the image is a bit different then the proposed by the task
- Icon size could be a little less agressive

PR #2909, opened 2025-12-27
add task solution

vectormeleshko:
https://vectormeleshko.github.io/ce-vzagali-jest


Anton-Kuchmasov (changes requested):
DEMO link doesn't work, provide correct link (re-deploy if needed) and re-request review again
image

PR #2917, opened 2026-01-03
add task solution

OlenaSobolieva:
task solution


etojeDenys (changes requested):
1. background image should fill the entire width of the screen
image

2. make sure the menu looks the same as the layout
image

3. disable page scrolling when the menu is open(point 11 in the [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md))
image

4. make sure these links work
image

5. these should be links as well
image

6. don't forget these steps
image

7. disable text area resizing
image

8. slightly increase the size of the images when hovering them

9. there is too much space here
image

etojeDenys (changes requested):
1. images should not overlap other items
image

2. this button is too big
image

3. this should be a button instead of a div
image

4. on the desktop version, phone numbers should appear when you hover this icon
image

example:
image

etojeDenys (approved):
good job

PR #2924, opened 2026-01-07
add task solution

Pedro031098:
[DEMO LINK](https://Pedro031098.github.io/layout_landing-page/).


IanLuan (changes requested):
You’re almost there! The images aren’t working on the DEMO LINK, and the container width should be 100%. On wider screens, it’s getting cropped.

raulriato (changes requested):
Something is not right. Your `DEMO LINK` is not up to date with your code. Your `pull request` is missing the `index.html`. Make sure to add the `index.html` to a commit and push it, so One of the mentors can review it properly

IanLuan (changes requested):
You’re almost there, but there are a few points to address:

1- The layout is still getting cropped on wider screens. You can reproduce this by zooming out the page. Make sure the header spans the full screen width, and apply max-width only to the inner content containers, centering them on the screen.
Image

2- The page shouldn’t reload on form submission. Instead, the fields should be cleared. You can use:

3- The grid layout is incorrect on wider screens. The images are appearing in the same row as the text. Use a grid to display the images, and place the paragraphs after the grid.
Image

4- All phone icons and phone numbers should be real links that initiate a call.

5- All addresses should be links to a location on Google Maps. Open them in a new tab using target="_blank".

6- The links inside the menu are not styled.

DanielCamposDevX (changes requested):
1. Numbers breaking page styling:

2.Grid layout breaking in bigger screens:

3. Form is not triggering a reload.

4. Footer image with wrong max width:
image

pedro-ruas (changes requested):
I still see some issues on the Demo Link:
- Links are not styled in the whole page
- The header has links showing that should not be there
- The contact form should have all fields set to `required`
- Images should have an smooth scaling animation on hover
- Page is a little to the left, not centralized

pedro-ruas (approved):

PR #2926, opened 2026-01-07
Develop

Teti-web:
Preview: https://teti-web.github.io/layout_landing-page/


vitalii-bondarenko-ua (changes requested):
Please, provide demo link to PR description

maxim2310 (commented):


Image

2. you can also add the same animation as for categories and make it clickable 

Image

3. this link has to be active

Image

4. prevent text area from being resized horizontally 

Image

5. all fields have to be required and submit button has to only reset form not the whole page 

Image

6. check padding at the bottom according to design 

Image

maxim2310 (changes requested):
.

2pasha (approved):
good job! 🔥

PR #2927, opened 2026-01-07
Develop

elenachernyshova01-cmd:
https://elenachernyshova01-cmd.github.io/layout_landing-page/


etojeDenys (changes requested):
demo link looks broken. please fix it and re-request the review

image

vitalii-bondarenko-ua (changes requested):
Good job.
Pay attention on [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md)
- Use the logo as a favicon
- Make sure phone icons and phone numbers are real links to start a call. Same for email
Screenshot 2026-01-08 at 16 15 53

- The page shouldn't be reloaded on form submit, but the fields should be cleared

Check all points from checklist

Also you need to fix horizontal scroll.
Screenshot 2026-01-08 at 16 11 35

https://github.com/user-attachments/assets/88c6e5e4-dd90-436d-b9c4-d65d078c7750

etojeDenys (changes requested):
almost done

1. add a hover effect to these links and ensure they lead to the correct section
image

2. on the desktop version, this button should be smaller
image

3. it should be tel link
image

4. images should be slightly bigger
image

5. phone number should only appear on the desktop version
image

Denys-Kravchuk9988 (changes requested):
Good job!

Please note that `h3` and `p` have default margins, which increases the spacing between the content.
image
image

I would recommend to use `padding-top` instead of `margin-top`. Also here the content slightly overflows the container and extends beyond its boundaries.
image

Here also `h3` has default margins
image

I would recommend to export `.png` images from design in 4x scale for better quality or download them in `.svg` or `.avif` format
image
image

I’d recommend making the button hover animation a bit smoother
image

etojeDenys (approved):
good job

PR #2951, opened 2026-01-21
Develop

itsmeoleg:
https://itsmeoleg.github.io/layout_landing-page/


etojeDenys (changes requested):
good job

1. add a transition to the hover effect because it is difficult to hover the phone number at the moment
image

2. phone icon should be a tel link (select one of the numbers)
image

3. decrease the scale of the hovered image to avoid overlap
image

4. add cursor pointer to the buttons

5. make sure that your buttons follow this layout
image

vitalii-bondarenko-ua (approved):
Looks good

PR #2962, opened 2026-01-27
add task solution

andreira-chilin:
https://andreira-chilin.github.io/layout_landing-page/


2pasha (changes requested):
good job! 👏

here are some improvements: 

1. these images should be in `svg` format
image

2. fix text in the first section
image

3. menu should be full viewport height
image

4. fix this button
image

andreira-chilin:
3. ok
but  height is right as like in the mockup
 don't think 700px is = 100 vh on desktop width as 1260 =(
 
Снимок экрана (1)

etojeDenys (changes requested):
almost done

1. decrease the scale of the hovered image to avoid overlap

Image

2. phone number should appear on the desktop version
Image

3. it would be better to set the height of this section to 100vh as well

Image

2pasha (approved):
LGTM 🔥

PR #2963, opened 2026-01-27
add task solution

GuiMachado001:
[DEMO LINK](https://GuiMachado001.github.io/layout_landing-page/).


raulriato (approved):
Very good!

Only tip for the next projectsis that ou've missed one item from the checklist.

But I'll approve, since it's not critical.

PR #2966, opened 2026-01-28
Develop -  Develop layout BO all done - checklist ok - Demonlink ok

Romulo-us-br:
Demonlink: https://Romulo-us-br.github.io/layout_landing-page/


DanielCamposDevX (changes requested):
You are almost there!! The background effect at the end of the page is a bit incorrect, only the image should be at -1.
Remember to chech font-sizing (some of then are a bit too small).
At the form remember too space the items from the edge.

pedro-ruas (approved):

PR #2985, opened 2026-02-08
add task solution

andre349:
- [DEMO LINK](https://andre349.github.io/layout_landing-page/)


2pasha (changes requested):
good job! 👏

here are some improvements: 

1. add favicon

Image

2. phone number has different font and remove bg from number

Image

3. slightly decrease image scaling on images to avoid overlapping 

Image

4. remove base outline

Image

5. placeholder's fonts should be the same over all inputs 

Image

6. add form validation
- form should be sent only if all fields are filled
- add email validation 

Image

7. here should be links to phone, mail and address 

Image

8. this block should have different gaps

Image

9. disable scrolling when menu is open 

Image

andre349:
Hello, I don’t understand how I should make it so that my main page doesn’t reload every time I submit a form. I think I did everything correctly, but where I have the button — it’s of type submit. Then, when the form is checked to see if all three fields are filled, and if the email is valid, it will submit. If I set the button type to reset, it won’t validate anything and will just reset the form. So I don’t understand how to do this without JS.

andre349:
Okay, everything seems to be working now — specifically, my page no longer reloads when the form is submitted.

2pasha (changes requested):
great progress 🚀 

here are last improvements:

1. your inputs and textarea have different fonts. Please use the same font

Image

Image

2. remove underlines

Image

etojeDenys (approved):
good job. just make sure the phone number only appears on the desktop version

PR #2986, opened 2026-02-08
add site Bikes

rabotadanyil:
https://rabotadanyil.github.io/layout_landing-page/


2pasha (changes requested):
good job! 👏

here are some improvements: 

1. first section should be full viewport height

Image

2. add hover effect on this icon 

Image

3. add hover effect for your items inside menu

Image

4. add hover effect for your images (slightly increase your image on hover)

Image

5. download image without watermark

Image

6. fix container in this section 

Image

7. add form validation
- form should be sent only if all fields are filled
- add email validation

Image

PR #2991, opened 2026-02-10
.

eduferrs:


pedro-ruas (changes requested):
There are 2 issues to solve:
- When hovering the phone icon at page top, the phone number should appear floating next to it
- All fields from the form should be required (the message too)
- When sending the form, the page is breaking. It should not send anything, just clear the form and not cause a reload

IanLuan (approved):
Great job!

PR #2994, opened 2026-02-11
landing-page

fmoreira85:


IanLuan (approved):
Great job! I’m accepting your task, but here’s a tip: currently, you’re using the mobile version of the header image on the desktop view, which makes it look low quality and cropped. Use the desktop version of the header image for larger screen sizes. Also, the container’s horizontal padding is 72px, but in Figma it’s 120px.

PR #3001, opened 2026-02-16
mensagem

JohannPDaniel:


IanLuan (changes requested):
The DEMO LINK is not working. Remember that you should modify the index.html file outside the src folder, not create a new one. Please fix this and run the deploy again.

PR #3009, opened 2026-02-21
Develop

aanimra:
- [DEMO LINK] (https://aanimra.github.io/layout_landing-page/).


etojeDenys (changes requested):
good job

1. on the desktop version, the phone number should appear when you hover that icon
Image

example: 

Image

2. make sure this button follows this mock-up

Image

mock-up:

Image

3. make sure the page looks correct in smaller versions(i think the problem is because of the buttons)
Image

Anton-Kuchmasov (changes requested):
When I'm chosing one of suggested input values, styles for [autofill ](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:autofill)should be the same - no blue background or other visual changes allowed:

Image

Would you mind to fix it?

Anton-Kuchmasov (approved):
Well done!

PR #3018, opened 2026-02-27
Develop

HakkeOzr:
[DEMO LINK](https://hakkeozr.github.io/layout_landing-page/)


Denys-Kravchuk9988 (changes requested):
Please add the [DEMO LINK](https://hakkeozr.github.io/layout_landing-page/). It seems that solution isn't ready yet

etojeDenys (changes requested):
1. increase the scale of images when hovering them. also use higher-quality images for the desktop version

2. that button is too big

Image

3. phone number should appear when you hover the phone icon
example: 
Image

4. make sure that your buttons follow the template

Image

5. the same applies to inputs and text area

Image

6. disable page scrolling when the menu is open (see [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md) point 11)
Image

etojeDenys (changes requested):
good job, to improve:
1. it should be tel, mailto, and address links

Image

2. disable text area resizing

Image

3. don't forget this step:

Image

4. add this hover effect to the desktop version
Image

5. use higher-quality images on the desktop version
Image

HakkeOzr:
.icon {
  display: block;
  height: 24px;
  width: 24px;
  background-position: center;
  background-size: cover;

  @include hover(transform, scale(1.2)); 

@mixin hover($property, $toValue) {
  transition: #{$property} $effectDuration;

  &:hover {
    #{$property}: $toValue;
  }
}

I have a hover effect on my computer for all versions.


(??????)

2pasha (changes requested):
good job! 👏

here are last improvements: 

Image

2. add hover effects for images like that

Image

brespect (changes requested):
Almost done, few things:
1. Change the favicon, for now i can't understand what is there.

2. Add hover to all interactive elements.

HakkeOzr:
Okay, I fixed the favicon, but regarding the second clarification: first of all, I’m not a designer or anyone involved in creative work at all; the Figma mockup doesn’t clearly indicate what exactly you want to see as an interactive element; Second, could you please specify exactly what you want? I have absolutely no idea what kind of interactivity is required for this task; Third, I would like my time to be valued as well, because there were several previous comments, and you could have specified this in them. P.S. Neither on Figma nor in the requirements was it specified exactly which favicon is needed, and this is an oversight on the part of your designers; I simply took the company name and added it.

2pasha (approved):
lgtm 🔥

about your previous comment: most of time you should add some basic things by yourself as a developer. For example: adding favicon, make sure that page title is correct, add `cursor: pointer` for all interactive element (buttons, cards, etc.) and so on. 

Because not only designer should think about UX, but you as a developer also should make sure that UX is good according to best dev practices. 

About reason why you had such a lot iteration, because in some cases fixing one problem - creates new problems and we can not predict all potential issues in advance.

HakkeOzr:
None of my edits broke the code, so new ones wouldn't appear. As for the design, these things aren't obvious to me - I paid for a full-stack solution, not for design. The person who sold me the course convinced me that I wouldn’t have to figure out what to add and where to add it myself. If you can simply add a cursor to all clickable links and you approve of this as part of the training program, I’ll add a cursor everywhere.

PR #3039, opened 2026-03-09
add task solution

IvanMaidanskyi:
[DEMO LINK] (https://IvanMaidanskyi.github.io/layout_landing-page/)


brespect (changes requested):
Good progress, come & check this: 
1. These icons looks too big:
Image

2: Add some hover effect to all this cards: 
Image

3. Add `cursor:pointer` to all buttons and icons:
Image

4. Add some underline with light effect on hover on this:
Image

5. Add `z-index` property to Footer:
Image

6. There should be a hover effect on phone icon:
Image

IvanMaidanskyi:
I did z-index property -1

brespect (changes requested):
Almost done check next:
1. Remove the outline on hover, let`s use scale or if you want make it light:
Image

2. Remove watermark from this image:
Image

3. The footer should be at the bottom of the page and create a slide-like effect.

2pasha (changes requested):
good job! 👏

here are some improvements: 

1. footer should not be here

Image

2. this field also should be required

Image

etojeDenys (approved):
lgtm

PR #3050, opened 2026-03-12
Moyo header solution

Palfinger2002:


ArturErmolenko (changes requested):
Plese add code and Demo link ;)

PR #3054, opened 2026-03-14
add task solution

chlodav:
https://chlodav.github.io/layout_landing-page/


etojeDenys (changes requested):
almost done

1. disable page scrolling when the menu is open (see [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md) point 11)

Image

2. make the button the same width as the fields

Image

3. phone numbers that appear when you hover the phone icon look broken
your: 

Image

example: 

Image

brespect (changes requested):
Good progress, check next:
1. This hover effect should has different positioning:
Image

2. The layout of menu header and side bar header should be the same for now I click the menu icon and after that I should move mouse to select close icon of side menu, that`s wrong positioning:
Image

3. Implement some hover effect for form fields.

4. Add favicon to browser tab.

etojeDenys (approved):
good job

PR #3055, opened 2026-03-15
add task solution

Alibobick:
-[DEMO LINK] (https://Alibobick.github.io/layout_landing-page/)


Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:
1. I would recommend to add a favicon
Image

2. For better quality of logos and icons I would recommend to export them in `.svg` from Figma 
Image
Image
Image

3. One image is broken here
Image

4. I would recommend to add `box-sizing: border-box;` for these buttons and inputs
Image
Image

5. Please, remove default button's borders or use anchor tag instead for buttons
Image

6. Address should be also a link and open google maps in a new tab with current location
Image

7. Here is needed to remove autofill styles
Image

8. It's better to disable scrolling when menu is open
Image

9. I would recommend to scale these images by hover and add `cursor: pointer;`
Image

Anton-Kuchmasov (changes requested):
Good job, but there are a few issues need to be fixed:


1. Clicking on `Send` button should erase all entered data - no scroll-ups or other side-effects. Check out #15 of [this instruction ](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md)to proceed:

Image

2. Would you mind to add some `hover` effects on each single link in the Contact Us section?
Also, pay attention: Google Maps should open in the new tab (`target: _blank`)

etojeDenys (approved):
good job

PR #3072, opened 2026-03-24
Develop

OlehYuriev:
[DEMO LINK](https://OlehYuriev.github.io/layout_landing-page/)


src/styles/blocks/compare.scss
```diff
@@ -0,0 +1,37 @@
+@use '../utils/variables' as v;
+@use '../utils/mixins.scss' as m;
```
dothearts, line 2: Understand that it v - variable, m - mixin, but for future better option, not use one letter naming

dothearts (changes requested):
In general i liked your solution. 
But you had some differences with the design (minority). 

1. Please fix full open menu, and header color. 
design - 

Image

solution - 

Image

2. Max scroll for footer image 
design - 

Image

solution -

Image

The image takes the whole screen if I scroll down, but in general, it needs to take only half of the screen, something like this 

Image
In this case, the form is left visible, and the photo takes only half of the screen. 

Fix these minority issues, and request review. 
Many thanks for your solution!

dothearts (approved):
Great job. 
Approved!

PR #3074, opened 2026-03-25
add task solution

timurradkevic:
- [DEMO LINK](https://timurradkevic.github.io/layout_landing-page/)


dothearts (changes requested):
Great work! 
I checked your layout and suggested changes for a better view. 

1. Fix the header color, and the hover bug with icons

Image

2. Add shifting effect for footer image 

Image

CheckList: 

Image

Many thanks for the solution, it's too close to design, only need to fix some small issues!

dothearts (changes requested):
Many thanks for the changes, but I still reproduce the same issues. 
Please, check the comments above. 

Also, you can check the footer scroll effect by [link](https://olehyuriev.github.io/layout_landing-page/#)

Make changes, request review! 
Thanks!

dothearts (approved):
Many thanks. 
Approved!

PR #3076, opened 2026-03-26
add solution

RuslanV23:
https://ruslanv23.github.io/layout_landing-page/


2pasha (changes requested):
good job! 👏

here are some improvements: 

1. first section should be full page height 

Image

2. use image without watermark 

Image

3. all fields should be required and add email validation 

Image

brespect (approved):
Well done! 👍

PR #3079, opened 2026-03-29
fix: manual stylelint cleanup

itNepes:
https://itNepes.github.io/layout_landing-page/


brespect (changes requested):
Good job, almost done, check next:
1. Add favicon to browser tab.
2. There must be the hover effect on phone icon:
Image

3. Remove watermark from this image:
Image

4. Image hover should not cut the space of another image.

5. Remove default focus styles and autocomplete from input fields:
Image

2pasha (changes requested):
good job! 👏

here are last improvements: 

Image

2. here should be link to google maps, exactly to that location

Image

3. container width should be the same for all sections 

Image

brespect (changes requested):
Almost done, few things:
1. This text should be different positioned:
Image

2. There should be the hover effect on phone icon, check the Figma Mockup:
Image

3. Gap between the form and info should be bigger:
Image

brespect (approved):
Well done! 🚀

PR #3083, opened 2026-03-29
landing-page

impo10nt:


dothearts (changes requested):
Check your solution and has some issues: 
In general, please check the design, section by section, because I see that some moments are skipped. 

1. Image position on first screen

Image

2. Hover effect

Image

3. Menu not in full screen

Image

4. Cards don't has hover effect 

Image

5. Off textarea transformation, user can do something like this: 

Image

6. In the footer, you need to make a scroll effect using z-index, not simply add an image 

Image

Image

[Example](https://sonik-boom71.github.io/layout_landing-page/) (check only footer effect) 

Please, check the design in detail, make changes, and then request review! 
Thanks!

src/styles/blocks/_navbar.scss
```diff
@@ -0,0 +1,118 @@
...
+  background-color: #c2c2c2;
+  color: $dark-text-color;
+
+  // @include on-tablet {
+  //   height: 630px;
+  // }
+
+  // @include on-desktop {
+  //   height: 712px;
+  // }
```
dothearts: Remove unused code

src/styles/blocks/_header.scss
```diff
@@ -0,0 +1,22 @@
...
+  &__menu {
+    display: flex;
+    gap: 24px;
+    height: 32px;
+
+    // &__content {
+    //   display: flex;
+    //   justify-content: space-between;
+    //   width: 100%;
+    // }
```
dothearts: remove unused code

src/scripts/main.js
dothearts, line 1: Also this file

dothearts (changes requested):
Great job!
 
Please update the first screen. 
Now on 2k monitor it looks like: 

Image

You can make a hero section with full width and height

design - 

Image

Please fix it, and the small comments that I left. 
Many thanks!

dothearts (approved):
Approved.

Many thanks!

PR #3106, opened 2026-04-02
add task solution

Iryna8989:
- Implemented landing page based on Figma design
- Added responsive layout
- Used SCSS and BEM

DEMO LINK: https://Iryna8989.github.io/layout_landing-page/


brespect (changes requested):
Good progress, check next:
1. There should be hover effect on phone icon:
Image

2. The gap between blocks of information is too big, check the mockup:
Image

3. There should be no horizontal scroll on Mobile:
Image

4. All this pictures on mobile should like square, not rectangles:
Image

Iryna8989:
Hi! Thanks for the feedback.

1 Added a hover effect on the phone icon — now the phone number appears on desktop.
2 Fixed the spacing between the information blocks according to the mockup.
3 The horizontal scroll on mobile was caused by a full-width image extending beyond the container. I fixed it by preventing horizontal overflow on the page.
4 Updated images on mobile to be square using aspect-ratio (in the video lesson it was done using the padding-bottom and absolute positioning , but I implemented it using aspect-ratio as a more modern approach)

2pasha (changes requested):
good job! 👏

here are last improvements: 

1. fonts should be the same

Image

2. some problem with header on mobile

Image

Iryna8989:
1.  Fixed. I missed the fact that form elements don't inherit the font-family by default
2. Fixed. The issue was caused by the phone number element taking up space on mobile even when invisible, which pushed the icons and caused layout issues. I solved it by setting display: none on mobile so it takes no space, and display: block on desktop with opacity: 0 and visibility: hidden by default which also allows a smooth hover transition on desktop

vadiimvooo (approved):
Well done

PR #3112, opened 2026-04-06
Develop

PatronXXi:
- [MYBIKE landing](https://www.figma.com/file/NZQAIydtHo5QkINyGLHNcq/BIKE-New-Version?node-id=0%3A1)
- [DEMO LINK](https://PatronXXi.github.io/layout_landing-page/).


Denys-Kravchuk9988 (changes requested):
Good job! 

A few things to improve:
1. The page is broken on screen width 1260px
Image
Image

2. The title text should be displayed on the center
Image

3. There should be phone number by hover on phone icon
Image
Image

4. I would recommend to remove autofill styles and add horizontal margins
Image

5. It's needed to export the image without discover the app icon
Image

6. It's needed to disable scrolling when menu is open
Image

7. It's better to align bike's banner according to design's banner
Image
Image

and the same for tablet
Image
Image

brespect (changes requested):
Good progress, check next:
1. Add the favicon to the browser tab.
2. Add hover effect to all interactive elements.
3. For now, I can't submit the form.
4. Add z-index property to make slide effect on Footer.
5. Textarea looks broken:
Image

2pasha (changes requested):
good job! 👏

here are some improvements: 

1. try this approach for form submitting 

Image

2. try `outline: none` to remove base input outline 

Image

3. here should be link to google maps

Image

4. disable form resizing like this 

Image

5. phone number should be appeared only on desktop 

Image

brespect (changes requested):
Almost done, fix this:
1. Main image should has different position on Mobile:
Image

2. Add some hover, transition/underline here:
Image

PatronXXi:
the height is too small, so the image tends to fill the space

Anton-Kuchmasov (changes requested):
Almost done!

1. Check out [CSS Autofill](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:autofill) styles here - make them the same as regular input (I mean, no `background-color` changed when autofill value entered is allowed):

Image

2. Click on `phone` icon (as well as on phone number beside) should trigger phone call. Use `<a>` with `tel` to proceed

Image

Anton-Kuchmasov (approved):
Amazing job!

PR #3123, opened 2026-04-08
add task solution

Jaseniin:
- [DEMO LINK] (https://Jaseniin.github.io/layout_landing-page/)


brespect (changes requested):
Good progress, check next:
1. There should be hover effect on phone icon:
Image

2. Add the animation to make open effect for side menu.

3. Add hover effect to all buttons.

2pasha (changes requested):
good job!

last fixes:

too big gaps

Image

Image

Jaseniin:
Should I also change the gap from the contacts section to the footer? I currently have min-height: 100vh; set there. 
When I switch it to margin-bottom, a black strip appears above the footer.

etojeDenys (approved):
good job, just remove this style

Image

PR #3143, opened 2026-04-23
Task solution

Rostyslav452:


2pasha (changes requested):
good job! 👏

here are some improvements: 

1. add [DEMO LINK](https://rostyslav452.github.io/layout_landing-page/) to PR's description

2. add favicon 

Image

3. this section looks broken 
- add content container
- text should be bigger and positioned little bit upper
- background image should be different

Image

4. add some transition for menu opening

5. when user hover on phone icon - phone number should be appear

Image

6. on desktop layout in this section should be different 

Image

7. this section looks interesting but please implement design from Figma

Image

8. on desktop here is different layout 

Image

9. fix form: 
- all fields should be required
- when user submit the form, you just need to reset form; no reload, no navigation

10. use footer for desktop

Image

11. please make sure that you have applied all changes before and check how it looks on tablet as well

etojeDenys (changes requested):
almost done

1. this link should open in a new tab

Image

2. remove base outline

Image

3. remove the autofill style

Image

4. all fields should be required
Image

5. phone number should only appear on the desktop version

Image

2pasha (changes requested):
good job! 👏

here are some improvements: 

1. all fields should be required and user should not be able to submit empty form 

Image

2. moments with autofill are still actual. You can see more [here](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:autofill)

Rostyslav452:
I have forgotten to deploy the new improvments)

etojeDenys (approved):
good job

PR #3147, opened 2026-04-27
first commit

Bo-dikkk:
https://Bo-dikkk.github.io/layout_landing-page/


brespect (changes requested):
Almost done, check next:
1. Add hover effect to all interactive elements.
2. Add the favicon.
3. Add correct grid-row value to the Contact Us form on Tablet.

etojeDenys (changes requested):
there are still some problems
1. slightly decrease the size of the hovered image to avoid overlaping

Image

2. it should be links

Image

3. add this hover effect to the desktop version

Image

4. disable page scrolling when the menu is open (see [checklist](https://github.com/mate-academy/layout_landing-page/blob/main/checklist.md), step 11)

Image

2pasha (changes requested):
good job! 👏

here are last improvements: 

1. this field also should be required 

Image

2. add `cursor: pointer` for all your interactive elements like this

Image

2pasha (approved):
lgtm 🔥

PR #3149, opened 2026-04-27
finished landing page

Dima23-ops5:


fetis-off:
I am wondering what is the task. If it's frontend course, then it appeared in list of Java homeworks by mistake.
I will reject this HW, because I cannot check it properly :) Please write in jv-chat about this task and we will manage this situation.

fetis-off (changes requested):
I am wondering what is the task. If it's frontend course, then it appeared in list of Java homeworks by mistake.
I will reject this HW, because I cannot check it properly :) Please write in jv-chat about this task and we will manage this situation.

Elena-Bruyako (changes requested):
the task isn't related to Java course, ask coordinators about the issue

PR #3158, opened 2026-05-02
landing page

kaida-iryna:


etojeDenys (changes requested):
demo link does not work. please fix it and request the review again

Anton-Kuchmasov (changes requested):
Good job!

To improve:

1. Reset all [CSS Autofill](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:autofill) styles to prevent UI diffs from regular input to autofill values:

Image

2. Also check out form styles on the screenshot above - button seems slightly wrong

3. Add link to Google Maps on the `Address` in the section on the right from input form

4. Add `hover` effect to all interactive elements on your page (links, buttons etc.)

5. Height of this section should be `100vh` on the PC view. Also check out `background-position` - it looks uncorrect now

Image

6. Links from `aside` menu should redirect user to the selected section

You can use [this solution](https://razor5000.github.io/my-bike-landing-page/) as a reference

vitalii-bondarenko-ua (approved):
Looks good

PR #3159, opened 2026-05-03
add solution

ALendyel-749:


Anton-Kuchmasov (changes requested):
DEMO LINK: https://alendyel-749.github.io/layout_landing-page/

You did a big amount of work, but there are still a lot of issues to fix:

1. Check out (whole) Figma design. There are a huge diffs between them and your implementation:

check out form styles on PC view
Image

 this section is so big - according to Figma mockup, it's height should be only `555px` on PC
Image

content section looks so thin here:
Image

check out twice `Take the Streets` header styles. Also, pay attention: on `hover` to the phone icon, phone number should appear (use link below to discover this effect):

Image

Also, keep in mind about `background-position` of the Header section - now it looks slightly wrong

You can use [this solution](https://razor5000.github.io/my-bike-landing-page/) as a reference

etojeDenys (changes requested):
good job, to improve:

1. disable page scrolling when the menu is open (see step 11 here)
Image

2. add a smooth scroll behaviour

3. it should link to the Google Maps address
Image

4. don't forget these steps

Image

vadiimvooo (changes requested):
Good job, but there are some things to improve:
1. Images page are missing
Image

2. On phone icon hover  phone number should be displayed (only on desktop layout)
Image

3. Logo is missing in menu page
Image

brespect (changes requested):
Make sure you have deployed your solution:
1. There are no images in your landing:
Image

2. Open & Close menu icons should be at the same position, for now it is different.

brespect (changes requested):
Hi, check my previous comment as nothing changes and looks broken

Denys-Kravchuk9988 (changes requested):
Good job!

1. Please, check the height of images and align them according to the design
Image
Image

2. `Explore` button should be centred and have less width according to design. Also check the textes above.
Image
Image

3. It's better to export this image without this badge
Image

4. I would recommend to remove autofill styles
Image

5. Phone icon is better to be hidden when menu is open
Image

6. It's better to align margins from borders according to the design. Paddings are 20px from borders.
Image

7. Part of content is hidden on menu
Image

8. The adaptation on tablet seems to be broken
Image

2pasha (changes requested):
good job! 👏

here are last improvements: 

1. add hover effect for your images 

Image

2. use image without watermark

Image

3. add hover effects for these items 

Image

brespect (approved):
Lgtm

PR #3162, opened 2026-05-09
task

khomenko-vlad:
https://khomenko-vlad.github.io/layout_landing-page/


Denys-Kravchuk9988 (changes requested):
Excellent work!

One tiny thing to improve:
1. It's better to remove autofill styles (background becomes bluer) from inputs
Image

vadiimvooo (approved):
Well done

PR #3173, opened 2026-05-14
develop

FS-Olexander:


2pasha (changes requested):
please deploy your app and add DEMO LINK to PR's description

etojeDenys (changes requested):
demo link does not work. please fix it and request the review again

if you get stuck on something, feel free to ask for help in the chat

etojeDenys (changes requested):
good, now we can see some progress. demo works, but the page does not see styles
Image

Anton-Kuchmasov (changes requested):
Good job!

To improve:

1. Add `cursor: pointer` to all interactive elements on your page (e.g., buttons, photos etc.)

2. Would you mind to add some transform/transition on all photos? `scale(1.04)` with smooth transition effect will be fine

Image

3. Remove all default `Css Autocomplete` styles from your form - no `background-color` or other diffs allowed even if user will pick up one of the predefined values:

Image

4. When aside menu is opened, you should remove an ability to scroll the page. Check out #11 of [this checklist](https://github.com/FS-Olexander/layout_landing-page/blob/main/checklist.md) to proceed

Image

Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve: 
1. It's better to add favicon
Image

2. I would recommend to remove default button's border width
Image
Image

3. It's better to add `cursor: pointer` by hover and these links should be interactive. Phone should open an alert to call, email should open Mail box and address should open Google maps
Image

4. It's better to export icons and logos from Figma in `.svg` format for better quality
Image
Image
Image

Anton-Kuchmasov (approved):
Well done!

PR #3180, opened 2026-05-20
fix

ageev2506:
- [x] Yellow stars are added with container modifier + pseudo-selector (NO extra classes)
- [x] Each BEM block has its own separate file
- [x] All Typical Mistakes from BEM lesson theory are checked.
- [x] Code follows all the Code Style Rules !
- [ ][DEMO LINK](https://ageev2506.github.io/layout_landing-page/).


brespect (changes requested):
DEMO LINK is not working:
Image

PR #3181, opened 2026-05-21
my landing page

mauroyuzo:
https://mauroyuzo.github.io/layout_landing-page/


raulriato (approved):
You did a good job! There's only one thing that could be better:

Ideally, the `header` of the page would fill the entire screen (The bike picture). You can see what I mean by that comparing your page with the `mockup`:

In your page the bike on the `header` is not complete in the picture.

Image

But in the `mockup` the picture show the the bike completely:

Image

But since that's just a tiny improvement, I'll leave it as a tip to do later on and approve your project!

PR #3185, opened 2026-05-24
solution

vladyslav-trundenko:
[DEMO LINK](https://vladyslav-trundenko.github.io/layout_landing-page/)


Anton-Kuchmasov (changes requested):
Hey, your solution looks unfinished. You should markup all the page, not only the first screen (header or hero section)

Pass through all lessons till the `Landing Page` module, and you'll fulfill your Landing page as it supposed to be
here's link where you can see how it might be: https://xeniakolomiiets.github.io/layout_landing-page/

Denys-Kravchuk9988 (changes requested):
Excellent job!

One thing I would recommend to complete:
1. It's better to remove autofill styles
Image

2. It's better to use `.svg` format for icons and logos for better quality
Image
Image

One recommendation:
I would recommend to align this banner according to design
Tablet: 
Image
Image

Phone: 
Image
Image

brespect (changes requested):
Almost done, but there should be logo on phone icon hover:
Image

brespect (approved):
Well done!

PR #3186, opened 2026-05-24
Develop

Nemarlin:
[DEMO LINK](https://Nemarlin.github.io/layout_landing-page/)


Denys-Kravchuk9988 (changes requested):
The styles aren't loaded. Please, check the configs and re-deploy the project

Image

The links seem to be broken
Image

Probably, the reason is that in field homepage there should link for your deploy. `"homepage": "https://nemarlin.github.io/layout_landing-page"`

Image


A few things to improve (what I can say right now):
1. I would recommend to export logos and icons in `.svg` format from Figma for better quality
Image

brespect (changes requested):
Almost done, check this:
1.  This hover should has different position, check the mockup:
Image

2. Looks broken:
Image

3. Close & Open menu icons should be at the same position:
Image

brespect (changes requested):
My previous comment about open & close menu icons was not fixed

brespect (changes requested):
Almost done, check this:
1. The gap between header and links should be different in side menu:
Image

2. Same way here, there should not be empty space:
Image

etojeDenys (approved):
good job

PR #3188, opened 2026-05-25
Develop

azesmmisha:
https://azesmmisha.github.io/layout_landing-page/


brespect (changes requested):
Good progress, check next:
1. This Image should be different on Desktop according to Figma Mockup:
Image

2. Add hover to all this refs:
Image

3. This Image should has different position on Mobile, according to Figma Mockup:
Image

4. No hover effect on phone icon:
Image

brespect (changes requested):
Almost done, you need to add this hover effect:
Image

vitalii-bondarenko-ua (approved):
Looks good.

PR #3191, opened 2026-05-27
add Landing Page

142536Def:
[DEMO LINK](https://142536Def.github.io/layout_landing-page/).
-[x] Check `background-image: url()` to be relative to the `main.scss`. So should start with `../images`.
-[x] If some points from the checklist are not relevant to your design, just skip it.


brespect (changes requested):
For now its only ready for mobile view, ask review again when you will be completely ready. Also I can't open side menu using burger icon:
Image

PR #3198, opened 2026-06-03
add task solution

Jonoreh:
[DEMO LINK](https://Jonoreh.github.io/layout_landing-page/)


Anton-Kuchmasov (changes requested):
Good job you did!

To fix:

1. Icons in the top of the header looks slightly bigger than expected:

Image

Image

2. Would you mind to reduce `scale` effect just a little bit to avoid overlapping?

Image

3. Add some paddings inside fields of your contact form (see Figma mockup):

Image

4. Would you mind to reset all [CSS Autofill](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:autofill) styles here?

Image

2pasha (changes requested):
good job! 👏

here are some more improvements: 

Image

2. fix this button on tablet

Image

3. here is too much space on mobile 

Image

2pasha (changes requested):
this should work only on desktop :)

Image

2pasha (approved):
lgtm 🔥

PR #3199, opened 2026-06-06
Develop

millionbuck1:
[- [DEMO LINK](https://millionbuck1.github.io/layout_landing-page/)


Denys-Kravchuk9988 (changes requested):
Good job! 

A few things to improve: 
1. By hover on logo left and right part are hidden
Image

2. It's better to remove autofill styles
Image

3. By hover the phone icon the phone number should be displayed.
Image
Image

4. I would recommend to have similar background color for the menu
Image

5. For better quality it's better to export logos and icons in `.svg` format
Image
Image
Image

6. Try to export this image without `Discover the app` badge
Image

7. It's better to disable scrolling when menu is open

8. I would recommend to display pictures according to design
Image
Image

brespect (changes requested):
Good progress, check next:
1. This hover effect should be on left side:
2. Burger Icon has different size from Phone Icon:
Image

3. Main image has different position on mobile:
Image

4. Add z-index property to Footer to make slide effect.

5. Add the favicon to browser tab.

6. Open & Close menu icons should be at the same spot:
Image

brespect (changes requested):
Good progress, but not all my previous comments were fixed:
1. This image should has different position, check the Figma:
Image

2. Open & Close menu icons has different positions, fix it, check the mockup:
Image
Image

2pasha (changes requested):
good job! 👏

here are some improvements: 

1. it will be better to have content container for all your sections (+header)

Image

2. will be better to position these icons on the same level 

Image
Image

3. phone number should be appeared only on desktop 

Image

PR #3201, opened 2026-06-06
add task solution

ggylyon:
[DEMO LINK](https://ggylyon.github.io/layout_landing-page/)


Moroz-Dmytro (changes requested):
1. When I hove over phone icon I should see phone number
2. When I click "Book a test ride" nothing happens
3. Add hover effects for icon in "Compare Bikes" section
4. When I rotate phone I'm not able to see all items in menu

Image

781calmstate (changes requested):
I don't think the menu should look like this for mobile phone:
Image

This background image doesn't look like it's supposed to on mobile phone according to design:

Image

Moroz-Dmytro (approved):

PR #3203, opened 2026-06-08
Implement mobile-first responsive B&O landing page layout

PavloPro7:
[DEMO LINK](https://PavloPro7.github.io/layout_landing-page/)


Anton-Kuchmasov (changes requested):
Good job!

To improve:

1. I suggest DON'T add `hover` effect on this section - as you can see, scaling this image causing horiziontal scroll, which is not allowed:

Image

2. Would you mind to reset [CSS Autofill](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:autofill) styles in your form?

Image

3. Add `cursor: pointer` to all interactive elements on your page (e.g. buttons, links)

4. Please reduce `scale` effect to avoid overlapping:

Image

etojeDenys (changes requested):
there are still the problems mentioned in the previous review. make sure you deploy your changes

also, to improve:
1. this number should be a tel link
Image

2. the phone number should only appear on the desktop version

Image

3. the background colours are different

Image

brespect (changes requested):
1. Almost done, but for now I can submit form without the message:
Image

2. Also, check the positining of this section, from Figma:
Знімок екрана 2026-06-09 о 09 06 06
And yours:
Знімок екрана 2026-06-09 о 09 06 54

brespect (approved):
Well done!

PR #3210, opened 2026-06-14
Develop

CesarCyber321:
[DEMO LINK](https://CesarCyber321.github.io/layout_landing-page/)
https://github.com/CesarCyber321/layout_landing-page


DanielCamposDevX (changes requested):
Some conditions were not met:

- Use the logo as a favicon
- All form fields should be required and have correct types (email, tel, etc.) Placeholders should give examples of what should be added, to make the expected format clear The page shouldn't be reloaded on form submit, but the fields should be cleared. Use:

DanielCamposDevX (approved):
Everything working, thanks fot attending to the bate-papo

PR #3212, opened 2026-06-15
need help

ProninaMariia:
https://ProninaMariia.github.io/layout_landing-page/


etojeDenys (changes requested):
good job, to improve:

1. it should link to the Google Maps address

Image

2. send button should clear the inputs without refreshing the page

Image

3. it looks broken. Instead of downloading that big photo, you should download each product photo separately

Image

4. these icons are invisible

Image

etojeDenys (changes requested):
1. use a higher-quality image for logo
Image

2. it looks good, but the images are cropped

Image

3. it should be an image and text

Image

example:
Image

4. don't forget this step

Image

roman-mirzoian (changes requested):
I see a new commit, but the demo link still points to an incomplete version. Make sure you've done a new deployment with the latest fixes:
Image

vadiimvooo (changes requested):
Good progress. There are still some fixes to do:
1. It should be possible to hover the mouse over the number. Right now it will just dissapear in case you try
Image
2. There are some issues with image positions and backgrounds. Please download proper one from the mockup and make a change
Image
3. There is the issue with position of about section in comparison with others
Image
4. The mobile version looks broken now.
Image

vitalii-bondarenko-ua (approved):
Good job

PR #3216, opened 2026-06-18
Develop

redfield-mp:
https://redfield-mp.github.io/layout_landing-page/


vitalii-bondarenko-ua (changes requested):
Provide, please, demo link to PR description.

2pasha (changes requested):
good job! 👏

here are some improvements: 

1. this section should be full device width 

Image

Image

3. add some transition for these hover effects 

Image

4. add some hover effects for these images as well 

Image

5. use image without watermark 

Image

6. disable page scrolling when menu is opened

2pasha (approved):
lgtm 🔥

PR #3225, opened 2026-06-22
added animation + fixed text and icons margins for tablet version

NazariiKaban:


etojeDenys (changes requested):
all good. keep going and re-request the review when you've finished

etojeDenys (changes requested):
great job! just a few small fixes:
1. use img without a mark

Image

2. it would be good to increase the size of the images when you hover them (but avoid overlaping or merging)

3. and you forgot last step here

Image

brespect (changes requested):
Good progress, check next:
1. This text should has different position on mobile:
Image

2. Add z-index property to Footer to make slide effect.

3. Add hover effect and cursor: pointer to all this links:
Image

4. Remove autofill styles from form fields:
Image

5. This section should has different height & blocks positions according to the mockup:
Image

2pasha (approved):
lgtm 🔥

PR #3232, opened 2026-07-02
Portfolio Project

carvalhoryahn-alt:
[DEMO] (https://carvalhoryahn-alt.github.io/layout_landing-page/)


DanielCamposDevX (changes requested):
- The sections should be the height off the screen.
- The page is not following the proper grid rules (layout is too close to the edges.)
- Disable page scroll when menu is open.
- The top of the page dont need to follow scroll

src/styles/blocks/mixins/mixin.scss
raulriato, line 1: `mixin` is not a block, it is an `utility`, as well as `variables`, so you should put both inside `src/styles/utils`, not inside `blocks`.

src/styles/blocks/container/container.scss
```diff
@@ -0,0 +1,154 @@
...
+      }
+    }
+
+    .products {
+      &__title {
+        grid-column: 1 / -1;
+        width: 280px;
+        font-size: 38px;
+      }
+    }
```
raulriato: You have styles for this class in two different files: here and `card.scss`. This is wrong, because neither represent this block. I'll give more information in the general review.

index.html
```diff
@@ -1,22 +1,343 @@
...
+    </script>
+
+</head>
+<!-- head region -->
+
+  <body class="page">
+
+     <!-- header region -->
+    <header class="header">
+      <div class="container">
```
raulriato: this is semanticly wrong. This `container` is not needed. The `header` is already the `container` of this section.

raulriato (changes requested):
your page is okay, but your code is not.

The `container` div you are using in some `sections` is wrong, you should not have it, the own `section` is the container on it's own.

Also the organization of files is not okay. You should organizes your styles like this:
```
styles/
├── blocks/
│   ├── about-us.scss
│   ├── categories.scss
│   ├── contact-us.scss
│   ├── header.scss
│   ├── index.scss (here is where you put the `@imports`
│   ├── main.scss (here you'll put the general styles, for body, etc.)
│   └── recommended.scss
└── utils/
    ├── mixins.scss
    └── variables.scss
```
The organization is very important because right now it's hard to even find improvement points on the code because I wasn't able to find all the code I would recommend you to change.

Overall your styles look fine, so I think after making the recommended changes your project will be good to go.

If you need any help, please join a `Bate papo com o mentor` meeting, or call for help in the [chat](https://mate.academy/pt-br/chats/39089)

DanielCamposDevX (changes requested):
Image

Page layout is still broken, did you forget to run deploy?
You neeed to correct the layout to ber as close as possiblee to figma.
Remember to check your dedmo link before re-submiting the review

DanielCamposDevX (approved):

PR #3246, opened 2026-07-08
add task solution

RomaPleshenko:
[DEMO LINK](https://RomaPleshenko.github.io/layout_landing-page/)


etojeDenys (changes requested):
demo link does not work, please fix it and re-request the review again

brespect (changes requested):
Good progress, check next: 
1. Main Image looks completely different on Mobile view from the mockup reference:
Image
Image

2. Check the Phone and Burger Menu icons, seem to be incorrect in size and different from mockup:
Image
Image

3. Add the favicon.

4. Both of header icons should have hover.

5. Add hover to all interactive elements: buttons, images, icons & links.

6. Got an error after form submit:
Image

7. Remove autofill styles from form inputs:
Image

8. Margin between this two sections in Contact Us is different from ther mockup:
Image

Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:
1. Logo should have bigger size on desktop according to design
Image

2. By hover the phone icon the phone number should appear
Image
Image

3. It's better to remove autofill styles by click on submit button. The form is better to be resetted and page shouldn't be reloaded.
Image

4. The address link should open Google Maps in a new tab
Image

5. It's better to add `cursor: pointer` by hover on all interactive items on the page
Image

6. It seems that adaptation is broken on tablet
Image
Image

7. I would recommend to disable scrolling when menu is open

8. The font-size should be bigger on tablet according to design
Image
Image

9. The form's adaptation looks broken on tablet
Image
Image

Anton-Kuchmasov (changes requested):
Good job, but there are a few issues need to be addressed:

1. `Header` section (with bike background-image) should have height: 100vh. Additionally, check out the `background-position` value - it seems to be slightly wrong

Image

2. Please remove all predefined [CSS Autofill](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:autofill) styles from your form - if User selects one of pre-filled values, there should be no background-color or other UI diffs:

Image

3. If User opens up your burger-menu and then close it, whole Page makes non-scrollable

Steps to reproduce:
1. Open `burger menu` via clicking on the burger menu icon
2. Close burger menu at any method up to you
3. Try to scroll page vertically


Would you mind to fix it?

Image

etojeDenys (approved):
lgtm

PR #3248, opened 2026-07-08
chore: update landing page demo and deploy settings

vaynexx:


dothearts (changes requested):
1. Remove textarea resize 

Image

2. Please share design link (in comments below), becouse i need check some moments, for example footer overflowing 

Image

3. Need be clickable and open mail 

Image

4. Move css variable separate (for font, color, bg, etc). Currenlty you duplicate this values, and can miss something in diff places. Use variables for it. 

In general great work. Give me design references, i check it in more detail.

PR #3250, opened 2026-07-11
solution: layout landing page

AlexanderOkhotnichenko:
[DEMO LINK](https://AlexanderOkhotnichenko.github.io/layout_landing-page/)


Anton-Kuchmasov (changes requested):
Good job!

To improve:
1. Click on each of those links should redirect (smooth behaviour) us to chosen section. Use anchor links with `id`s to proceed

Image

2. Would you mind to add some transition to open/close actions of Burger menu? Now it affects immediately, which is not best option to UX

3. Add some `hover` effect on the product cards. `transform: scale(1.04)` with smooth effect will be the fine choice

Image

4. Please reset CSS Autofill styles - your main goal is to get rid all of them (when user selects one of pre-defined options, no `background-color` or other UI diffs allowed)

Image

5. Add links & hover effects to the `Contact Us` contacts

etojeDenys (approved):
great job

PR #3251, opened 2026-07-12
Develop

MasyaArc:
[DEMO LINK](https://MasyaArc.github.io/layout_landing-page/).


Anton-Kuchmasov (changes requested):
Good job!

To improve:
1. Phone number should be visible ONLY in `hover` state of phone icon:

Image

2. Would you mind to reset all [CSS Autofill](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:autofill) styles inside the form inputs?
I mean, when User select one of predefined values, there should be no `background-color` or other UI diffs - the same as regular input

Image

brespect (approved):
Looks good!

PR #3257, opened 2026-07-15
Develop_illia_amyk

illyaSmyk:
[DEMO LINK](https://illyaSmyk.github.io/layout_landing-page/)


roman-mirzoian (changes requested):
The code looks great, but please add a link in the description to the deployed result so it can be visually verified

roman-mirzoian (changes requested):
Good job!

Just a few things to improve:
1. Add some `hover` effect on the product cards. `transform: scale(1.04)` with smooth effect will be the fine choice.
2. The phone icon in the header should display the phone number when hovered over And when you click on it, it works the same way as elsewhere: the phone number pops up.:
https://www.figma.com/design/DtkQmQ797hk0nI4KfMi2Uq/BOSE-New-Version?node-id=6817-462&t=RQVfIwpwlOSL0s0y-0
3. You need to add the appropriate attribute to the address so that clicking on it opens a map showing that address.
Image

2pasha (changes requested):
good job! 👏

here are some improvements: 

1. disable scrolling when menu is opened
2. add favicon 
3. add hover effects for these images as well 

Image

4. all fields should be required and after form submitting page should not be scrolled to the top

Image

roman-mirzoian (approved):
Well done!

PR #3261, opened 2026-07-16
add task solution

ILLYA-M:


2pasha (changes requested):
good job! 👏

here are some improvements: 

1. add [DEMO LINK](https://illya-m.github.io/layout_landing-page/) to PR desc
2. disable page scrolling when menu is opened 

Image

3. slightly decrease hover effect scale to avoid overlap like these

Image
Image

4. here should be links to number, email and google maps 

Image

5. fix form:
- all fields should be required
- add email validation
- after form submitting - form should should be reseted, without scrolling, navigation and so on

Image

6. make sure that mobile hero section looks exactly as in Figma 

Image

7. phone number appearing should be only on desktop 

Image

8. also here should be link to call the number 

Image

Anton-Kuchmasov (changes requested):
Great job, but there are a few issues need to be addressed:

1. Remove all [CSS Autofill](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:autofill) styles inside your Form - basically, there should be no UI diffs between regular input & choosing from predefined values:

Image

2. Lock out `resizing` of your textarea - otherwise, we could have this effect:

Image

3. Background positioning looks slightly wrong:

Image

Take a quick look on [this link](https://extymandriy.github.io/layout_landing-page/) - you can use it as a reference and observe all needed styles with devTools

Anton-Kuchmasov (changes requested):
Issues from prev review weren't fixed:

Image

Image

vitalii-bondarenko-ua (approved):
Looks good

PR #3269, opened 2026-07-21
finish the landing page

justnicko:


roman-mirzoian (changes requested):
That's a great start! 
Here's what needs to be improved:
1. When you hover over the phone icon, the phone number itself should appear
Image

2. The main page title should be larger and positioned as shown in the design.
By design:
Image
Current version:
Image

3. The images on the page should be free of unnecessary elements.
Image

4. All form fields should be required
5. The image in the page footer is missing:
Image

2pasha (approved):
lgtm 🔥

PR #3273, opened 2026-07-23
add

miabellaa:
https://miabellaa.github.io/layout_landing-page/


2pasha (changes requested):
good job! 👏

here are some improvements: 

1. first section should be full page heigh 

Image

2. add favicon 

Image

3. when user hover on phone icon - phone number should be appeared

Image

4. disable page scrolling when menu is opened
5. it's better to use these images in `svg` format to increase quality 

Image

6. add hover effects for these images 

Image

7. here should be links to google maps 

Image

8. after form submitting page should not be reloaded. Form just should be reseted without reload, navigation or smth like that

roman-mirzoian (approved):
Well done 👍

PR #3274, opened 2026-07-23
created landing page

vozniyroman8-netizen:


etojeDenys (changes requested):
almost done
1. this link should open in a new tab

Image

2. make sure that this field requires an email address

Image

3. don't forget to implement this hover effect on the desktop version

Image

roman-mirzoian (changes requested):
Good work! All that’s left is to fix the phone element located in the header next to the menu: when hovered over, it should display the phone number (as noted in point 3 of the previous review), and when clicked, it should function the same way as other areas with phone numbers (currently, clicking it takes you to the landing page).

Image

Also, please don't forget to include a link to your landing page demo in the comment section of the PR.

etojeDenys (changes requested):
there are still problems that I mentioned in my previous review. make sure you deploy the changes

etojeDenys (approved):
lgtm

PR #3277, opened 2026-07-25
Develop

smarticboy731-hub:
https://smarticboy731-hub.github.io/layout_landing-page/


roman-mirzoian (changes requested):
Great start! Here are the issues that need to be worked on:
- Make sure phone icons and phone numbers are real links to start a call (the bottom icon is a link, but not the phone numbers)
- All addresses should be links to some location on Google Maps. Open them in a new tab with target="_blank"
- Disable page scrolling under the menu 
- The font size of the section headings should be adjusted to match the dimensions specified in the design
- The “View All” button should be designed as shown in the mockup
- For placeholders in the inputs, padding must be added everywhere
- For the tablet version, the images need to be centered correctly

etojeDenys (changes requested):
demo link does not work. please fix it and re-request the review

2pasha (approved):
lgtm 🔥

PR #3283, opened 2026-07-28
add solution

lubovmrk:
[DEMO LINK](https://lubovmrk.github.io/layout_landing-page/)


brespect (changes requested):
Good progress, check next:
1. Add the favicon.

2. There should be hover effect on Phone Icon:
Image

3. Remove autocomplete and cursor should has left padding:
Image

4. Remove resize ability and add top padding to placeholder:
Image

5. All these links in Footer should have valid reference and hover:
Image

roman-mirzoian (changes requested):
Great job!
There are just a few points left to correct:
- All logos should be links to the top of the page (even in Recommended section)
- Disable page scrolling under the menu
- The page shouldn't be reloaded on form submit
- The hover effect on the inputs is broken right now, it would be better to remove it. Also,  it should be set to `cursor: pointer` on all buttons:
https://gyazo.com/409cf040a840b5ea36bc95756b13a493
- There should be hover effect on Phone Icon:
Image

roman-mirzoian (changes requested):
Great job!
There's just one more thing to fix:
```html
```

roman-mirzoian (approved):
Well done 👍

PR #3287, opened 2026-07-29
Add MYBIKE landing page

HiljSV:
DEMO LINK: https://hiljsv.github.io/layout_landing-page/

  Checklist:
  - [x] Favicon uses logo
  - [x] Page title is MYBIKE
  - [x] Logos link to top
  - [x] Phone, email, address hover states work
  - [x] Phone links use tel:
  - [x] Address opens Google Maps in a new tab
  - [x] Images scale on hover
  - [x] No horizontal scroll on mobile
  - [x] Smooth navigation links
  - [x] Menu scroll lock and sticky close button work
  - [x] Form fields are required and use correct types
  - [x] Form clears without page reload
  - [x] Sticky footer effect is implemented


css/main.css
```diff
@@ -0,0 +1,797 @@
+*,
```
rostyslav-horytskyi: This file looks like compiled CSS, you don't need to commit it. Parcel builds it automatically when you run `npm run deploy`. Just remove this file

rostyslav-horytskyi (approved):
Good job! Keep it going!👏

Don't forget to fix comment🙂

css/main.css
HiljSV: Thank you for the comment. I removed css/main.css from the repository because it was a compiled CSS file and is not needed in the source code.
The project builds styles from src/styles/main.scss, and the generated CSS is created during build/deploy.

PR #3289, opened 2026-07-29
add task solution

Artroller:
- [DEMO LINK](https://Artroller.github.io/layout_landing-page/).


2pasha (changes requested):
good job! 👏

here are some improvements: 

1. change favicon 

Image

Image

3. looks like you are using wrong fonts 

Image

4. add hover effects for your images 

Image

5. implement hover effect for your buttons 

Image

6. container should be in this section 

Image

7. in address should be links to google maps. Also add hover effects for these links

Image

8. fix form 
- all fields are required
- add email validation 
- form should be reseted after submit (no refresh, on navigation etc)

9. disable scrolling when menu is opened 

10. fix mobile view 

Image

11. on mobile these images should be squares

Image

